PR #2, opened 2022-03-16
Add content flags

LornaHarwood:
Added content flags for:

- Case contains sensitive content
- Supplemental documents available

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog 1.0.0].
 
 ## [Unreleased]
 
+## [Release 2.0.0]
+- Refactored property accessor methods
+- BREAKING CHANGE `is_document_published` changed to `get_published` and `publish_document` changed to `set_published`.
+
 ## [Release 1.0.5]
 - Initial tagged release
 
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -208,12 +208,12 @@ class MarklogicApiClient:
     def set_boolean_property(self, judgment_uri, name, value):
         uri = f"/{judgment_uri.lstrip('/')}.xml"
         xquery_path = os.path.join(
-            ROOT_DIR, "xquery", "{name}.xqy"
+            ROOT_DIR, "xquery", "set_property.xqy"
         )
         string_value = "true" if value else "false"
         return self.eval(
             xquery_path,
-            vars=f'{{"uri":"{uri}","{name}":"{string_value}"}}',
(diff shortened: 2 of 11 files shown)


src/caselawclient/Client.py
```diff
@@ -219,6 +219,58 @@ def is_document_published(self, judgment_uri):
...
+
+        response = self.eval(
+            xquery_path, vars=f'{{"uri":"{uri}"}}', accept_header="multipart/mixed"
+        )
+
+        content = decoder.MultipartDecoder.from_response(response).parts[0].text
+        xml = etree.fromstring(content)
+        return xml.text == "true"
+
+    def supplemental_document(self, judgment_uri, supplemental=False):
```
Floppy, line 248: These methods are very similar, it would be good practice to reduce the duplication by combining them. You could create an method that all three of them call, which accepts an xquery filename and the other appropriate params. Same for the `is_document_*` methods too.

src/caselawclient/xquery/is-sensitive.xqy
```diff
@@ -0,0 +1,7 @@
+xquery version "1.0-ml";
+
+declare variable $uri as xs:string? external;
+
+let $prop := fn:QName("","sensitive")
```
Floppy, line 5: the prop name here could be an external variable passed in, then you wouldn't need three almost-identical xquery files.

src/caselawclient/xquery/sensitive.xqy
```diff
@@ -0,0 +1,8 @@
+xquery version "1.0-ml";
+
+declare variable $uri as xs:string? external;
+declare variable $sensitive as xs:string? external;
+
+let $props := ( <sensitive>{xs:boolean($sensitive)}</sensitive> )
```
Floppy, line 6: Same here with the external variable, but I'm not sure if that plays nicely with the element tag in this one? Worth a go though.

Floppy (changes requested):
Looks like it will work fine, but it would be good to reduce the repetition by extracting shared functionality into another method.

lozette:
You'll also need to increment the version number in `setup.cfg` otherwise the package won't publish! (As I have found out!)

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog 1.0.0].
 
 ## [Unreleased]
 
+## [Release 2.0.0]
+- Refactored property accessor methods
+- BREAKING CHANGE `is_document_published` changed to `get_published` and `publish_document` changed to `set_published`.
```
lozette, line 11: Nicely pointed out!

lozette (approved):
This is great 👏

Floppy (approved):

PR #8, opened 2022-03-18
Add testing to API client

LewisDaleUK:


tests/test_client.py
```diff
@@ -115,8 +115,6 @@ def test_set_boolean_property_false(self):
                 accept_header="application/xml"
             )
 
-
-
```
lozette: Can you squash this commit into 786a67802c337a06252c8bc1b56dcd6e484ee3b1

lozette (changes requested):
One small change please!

tests/test_client.py
LewisDaleUK: Done!

lozette (approved):

PR #16, opened 2022-03-31
Replace lxml with xml.etree from the Python standard library

LewisDaleUK:


lozette (approved):
Can you update the commit message with a quick explainer on why this is better!

PR #35, opened 2022-04-21
Fix/send default elements if they are missing

lozette:
If Elements are missing from the XML which we need to edit (e.g
if we are attempting to edit the metadata name of a judgment but its
`akn:FRBRWork` element is completely missing) then send an empty Element
in its place instead of sending an error.

This will allow editors to correct the missing metadata in the editor UI.

Rollbar: https://rollbar.com/dxw/tna-caselaw-editor-ui/items/14/

### DIFF
--- a/src/caselawclient/xml_tools.py
+++ b/src/caselawclient/xml_tools.py
@@ -3,6 +3,8 @@ from xml.etree.ElementTree import Element
 
 akn_uk_namespaces = {"akn": "http://docs.oasis-open.org/legaldocml/ns/akn/3.0",
                      "uk": "https://caselaw.nationalarchives.gov.uk/akn"}
+akn_namespace_uri = "http://docs.oasis-open.org/legaldocml/ns/akn/3.0"
+uk_namespace_uri = "https://caselaw.nationalarchives.gov.uk/akn"
 search_namespace = {"search": "http://marklogic.com/appservices/search"}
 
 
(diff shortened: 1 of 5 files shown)


src/caselawclient/xml_tools.py
```diff
@@ -3,26 +3,30 @@
...
-def get_element(xml: ElementTree, xpath) -> Element:
+def get_element(xml: ElementTree, xpath, element_name="FRBRname", element_namespace=akn_namespace_uri, has_value_attribute=True) -> Element:
     name = xml.find(
         xpath,
         namespaces=akn_uk_namespaces,
     )
 
     if name is None:
-        raise JudgmentMissingMetadataError
+        element = ElementTree.Element(ElementTree.QName(element_namespace, element_name))
```
Floppy, line 26: I see that this creates an element if the element doesn't exist, but is it created at the right point in the tree? If the value of this is set, will it actually be saved back into the XML document properly? Feels like it's just a floating element, unattached to `xml`?

lozette: Right, this is definitely the case *but* we can find the element's parent in the editor UI project and insert it into the parent. This works! (Tried it!). It's not very elegant though.

Floppy (approved):
Approved on the basis that the editor currently adds the element into the tree (see https://github.com/nationalarchives/ds-caselaw-editor-ui/pull/98), though I think in the long run it would be neater done in here.

PR #41, opened 2022-04-29
Fix: If metadata name element does not exist, insert a new one as a child

lozette:
We were assuming that the `akn:FRBRname` node would exist and could be edited,
but in some cases this node does not exist if the parser was unable to find a
case name in the judgment.

Check to see if a `akn:FRBRname` node exists in the document, and if it does
replace it, otherwise create a new one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog 1.0.0].
 
 ## [Unreleased]
+- Create a new `akn:FRBRname` node for the judgment metadata name, if one does not exist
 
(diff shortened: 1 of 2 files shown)


Floppy (approved):
This looks fine, but didn't we fix this in the 4.4.0 release and the editor? Is this another method by which things get set that that fix didn't affect?

lozette:
We did, but in that one we were returning a node to the editor, it was being edited and saved on the editor side. This was causing the namespaces to be wiped out. Everything is now being amended in ML via XQueries, but when I released 4.5.1 I forgot that sometimes the elements don't exist 🤦 

We can probably strip out the XMLTools functions which return elements, eventually.

PR #50, opened 2022-06-20
Move/rename judgment at URI

lozette:
## Changes in this PR:

An XQuery & method to rename/move a judgment from one URI to another.

Trello: https://trello.com/c/8n1Jks1r


src/caselawclient/xquery/move_judgment.xqy
```diff
@@ -0,0 +1,10 @@
+xquery version "1.0-ml";
+
+import module namespace dls = "http://marklogic.com/xdmp/dls" at "/MarkLogic/dls.xqy";
+
+declare variable $old_uri as xs:string external;
+declare variable $new_uri as xs:string external;
+
+let $judgment := fn:doc($old_uri)
+
+return dls:document-insert-and-manage($new_uri, fn:true(), $judgment)
```
dragon-dxw, line 10: document-insert sounds like a create rather than a transform -- does this leave the document at the old URI?

Docs at https://docs.marklogic.com/dls:document-insert-and-manage feel like they agree that it's a new doc.

dragon-dxw (approved):

tests/test_client.py
```diff
@@ -356,4 +356,22 @@ def test_delete_document(self):
                 os.path.join(ROOT_DIR, "xquery", "delete_judgment.xqy"),
                 vars=json.dumps(expected_vars),
                 accept_header="application/xml"
-            )
\ No newline at end of file
+            )
+
+    def test_move_judgment(self):
```
dragon-dxw, line 361: This function is still misnamed :)

PR #55, opened 2022-07-11
Set the judgment's internal URI (`FRBRthis` and `FRBRuri` nodes)

lozette:
## Changes in this PR:

In https://github.com/nationalarchives/ds-caselaw-public-access-service/blob/main/marklogic/src/main/ml-modules/root/judgments/xslts/judgment2.xsl#L69
we can see that the XSLT uses the node at `/akomaNtoso/judgment/meta/identification/FRBRWork/FRBRthis/https://github.com/value`
to determine the URI of a document, and then uses that URI to build up a URL
for the judgment's images.

In judgments which have failed processing, and have had a Neutral Citation Number
(and therefore a URI) added manually by the editors, this internal URI at
`/FRBRWork/FRBRthis/https://github.com/value` remains blank.

This has the knock on effect of image URLs not resolving correctly in the HTML
version of a judgment (which is generated from the above XSLT).

This commit introduces an XQuery & API endpoint to set the value of
`FRBRWork/FRBRthis/https://github.com/value`

This can then be used in the Editor UI to set the internal URI of a judgment
once a judgment is edited by the editors.

Additionally, the values of FRBRExpression/FRBRthis/https://github.com/value and FRBRExpression/FRBRuri/@value
should be set to the URI (but without the `/id/`), and the values of
FRBRManifestation/FRBRthis/@value and FRBRManifestation/FRBRuri/@value should be set
to the URI, without the `/id/` and with `data.xml` on the end.

E.g.

```
<identification source="#tna">
  <FRBRWork>
    <FRBRthis value="https://caselaw.nationalarchives.gov.uk/id/ewhc/scco/2022/1148" />
    <FRBRuri value="https://caselaw.nationalarchives.gov.uk/id/ewhc/scco/2022/1148" />
    <FRBRdate date="2022-05-04" name="judgment" />
    <FRBRauthor href="#ewhc-seniorcourtscosts" />
    <FRBRcountry value="GB-UKM" />
    <FRBRnumber value="1148" />
  </FRBRWork>
  <FRBRExpression>
    <FRBRthis value="https://caselaw.nationalarchives.gov.uk/ewhc/scco/2022/1148" />
    <FRBRuri value="https://caselaw.nationalarchives.gov.uk/ewhc/scco/2022/1148" />
    <FRBRdate date="2022-05-04" name="judgment" />
    <FRBRauthor href="#ewhc-seniorcourtscosts" />
    <FRBRlanguage language="eng" />
  </FRBRExpression>
  <FRBRManifestation>
    <FRBRthis value="https://caselaw.nationalarchives.gov.uk/ewhc/scco/2022/1148/data.xml" />
    <FRBRuri value="https://caselaw.nationalarchives.gov.uk/ewhc/scco/2022/1148/data.xml" />
    <FRBRdate date="2022-06-16T00:40:44" name="transform" />
    <FRBRauthor href="#tna" />
    <FRBRformat value="application/xml" />
  </FRBRManifestation>
</identification>
```

### DIFF
--- a/src/caselawclient/xquery/set_metadata_this_uri.xqy
+++ b/src/caselawclient/xquery/set_metadata_this_uri.xqy
@@ -1,0 +1,34 @@
+xquery version "1.0-ml";
+
+declare namespace akn = "http://docs.oasis-open.org/legaldocml/ns/akn/3.0";
+declare variable $uri as xs:string external;
+declare variable $content_with_id as xs:string external;
+declare variable $content_without_id as xs:string external;
+declare variable $content_with_xml as xs:string external;
+
+(
+  xdmp:node-replace(
+    document($uri)/akn:akomaNtoso/akn:judgment/akn:meta/akn:identification/akn:FRBRWork/akn:FRBRuri,
+    <akn:FRBRuri value="{$content_with_id}" xmlns="http://docs.oasis-open.org/legaldocml/ns/akn/3.0"/>
+    ),
+  xdmp:node-replace(
+    document($uri)/akn:akomaNtoso/akn:judgment/akn:meta/akn:identification/akn:FRBRWork/akn:FRBRthis,
+    <akn:FRBRthis value="{$content_with_id}" xmlns="http://docs.oasis-open.org/legaldocml/ns/akn/3.0"/>
+    ),
+  xdmp:node-replace(
+    document($uri)/akn:akomaNtoso/akn:judgment/akn:meta/akn:identification/akn:FRBRExpression/akn:FRBRuri,
+    <akn:FRBRthis value="{$content_without_id}" xmlns="http://docs.oasis-open.org/legaldocml/ns/akn/3.0"/>
+    ),
+  xdmp:node-replace(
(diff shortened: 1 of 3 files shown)


src/caselawclient/xquery/set_metadata_this_uri.xqy
```diff
@@ -0,0 +1,15 @@
+xquery version "1.0-ml";
+
+declare namespace akn = "http://docs.oasis-open.org/legaldocml/ns/akn/3.0";
+declare variable $uri as xs:string external;
+declare variable $content as xs:string external;
+
+(xdmp:node-replace(
+    document($uri)/akn:akomaNtoso/akn:judgment/akn:meta/akn:identification/akn:FRBRWork/akn:FRBRuri,
```
dragon-dxw, line 8: Do the Manifestation and Expression need to match, as well as the Work? (At this time, I think so.)

Somewhat concerned that this is quite a bit of editing in a different place -- is there someway we can make the Transformation Engine responsible for all this?

dragon-dxw (approved):

PR #62, opened 2022-08-05
Respect & maintain the `FRBRdate` attribute `name` in `FRBRWork`

lozette:
## Changes in this PR:

Related to https://github.com/nationalarchives/ds-caselaw-public-ui/pull/316

This XQuery sets or updates the node at `/akn:akomaNtoso/akn:judgment/akn:meta/akn:identification/akn:FRBRWork/akn:FRBRdate`

However, it does not respect the value of the `@name` attribute. It is being
set to `judgment`, which is true in many cases, but may also be `hearing` in
some cases.

In the referred PR above, we changed the way we get a judgment's date, from
looking for an `FRBRdate` with an attribute of `judgment`, we instead
look for any `FRBRdate` inside the `FRBRWork` element. So the attribute `name`
value is redundant.

When updating a judgement's date, we do not want to obliterate the existing
`name` attribute value, and replace it with `judgment`, as we were previously
doing. Instead we want to preserve it in the new/updated `FRBRWork/FRBRdate`
element.

However, if a judgment does not have an `FRBRWork/FRBRdate` element, create
a new one and give it the `@name` attribute `judgment` by default.

https://trello.com/c/lTTnAqpz/847-date-editor-ui

### DIFF
--- a/src/caselawclient/xquery/set_metadata_date.xqy
+++ b/src/caselawclient/xquery/set_metadata_date.xqy
@@ -4,13 +4,20 @@ declare namespace akn = "http://docs.oasis-open.org/legaldocml/ns/akn/3.0";
 declare variable $uri as xs:string external;
 declare variable $content as xs:string external;
 
+(: If at least one FRBRWork/FRBRdate element exists then...:)
 if (fn:boolean(
-cts:search(doc($uri),
+cts:search(doc($uri)/akn:akomaNtoso/akn:judgment/akn:meta/akn:identification/akn:FRBRWork,
 cts:element-query(xs:QName('akn:FRBRdate'),cts:and-query(()))))) then
-    xdmp:node-replace(
-    document($uri)/akn:akomaNtoso/akn:judgment/akn:meta/akn:identification/akn:FRBRWork/akn:FRBRdate,
-    <akn:FRBRdate date="{$content}" name="judgment" xmlns="http://docs.oasis-open.org/legaldocml/ns/akn/3.0"/>)
+    (: ...iterate over the FRBRdate element(s) :)
+    for $frbr_date in doc($uri)/akn:akomaNtoso/akn:judgment/akn:meta/akn:identification/akn:FRBRWork/akn:FRBRdate
+        (: Store the existing @name attribute value :)
+        let $attribute_name := $frbr_date/@name
+        (: Replace the exisitng FRBRdate element with a new one, using the same @name attribute :)
+        return xdmp:node-replace(
+        document($uri)/akn:akomaNtoso/akn:judgment/akn:meta/akn:identification/akn:FRBRWork/akn:FRBRdate[@name = $attribute_name],
+        <akn:FRBRdate date="{$content}" name="{$attribute_name}" xmlns="http://docs.oasis-open.org/legaldocml/ns/akn/3.0"/>)
 else
+    (: No FRBRdate exists in FRBRWork, so create a new one and use the 'judgment' attribute name as default  :)
     xdmp:node-insert-child(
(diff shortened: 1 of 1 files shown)


erbridge (commented):
Are there any tests to cover this? Not sure I know enough XQuery to know what this is doing.

dragon-dxw:
I'd be happier if it was more obvious that we were looking at the same object, rather than referencing it three times, by saying something like:

`let $relevantWork := document($uri)/akn:akomaNtoso/akn:judgment/akn:meta/akn:identification/akn:FRBRWork`

and then using `$relevantWork/akn:FRBRdate/@name` etc later.

I think there can be multiple `FRBRdates` in a `FRBRWork` meaning `$attributename` could be a list of `@name`s, which makes everything super complicated.

And I don't know that 'set metadata date' makes sense as an operation when there could be multiple metadata dates, and we replace all of them with a single date, when the hearing and judgment presumably occured on different dates?

(F's testing question is a good one too.)

lozette:
The testing question i a good one but I'm not aware of any mature testing processes for XQuery! There is https://dev.w3.org/2006/xquery-test-suite/PublicPagesStagingArea/ but it's not a mature thing and I don't know if I want to add something so experimental when we're still relatively pushed for new features?

erbridge:
Maybe integration level tests? I assume this gets run by some Python. Can we test that does what we expect?

lozette:
If this is something you think we need as a matter of urgency I can ask TNA for the time to figure out how we can test this at an integration level. The client is run in Python but I don't know off-hand how to mock a Marklogic database and see that the query runs and modifies the XML as expected. There are unit testing frameworks for Marklogic but they're all Java (that I can find).

lozette:
Unfortunately something like `let $frbr_work := doc($uri)/akn:akomaNtoso/akn:judgment/akn:meta/akn:identification/akn:FRBRWork` is an unsearchable expression and can't be used in `cts:query` (unless I'm missing something).

Agree that there could be multiple dates in an FRBRWork - I will check with Jim what we should do if there are multiple.

lozette:
Updated the XQuery to support multiple `FRBRWork/FRBRdate` elements.

Also have corrected a mistake - if no `FRBRWork/FRBRdate` exists, use `judgment` as the default `@name` attribute when creating a new element.

src/caselawclient/xquery/set_metadata_date.xqy
```diff
@@ -4,13 +4,20 @@ declare namespace akn = "http://docs.oasis-open.org/legaldocml/ns/akn/3.0";
...
-    xdmp:node-replace(
-    document($uri)/akn:akomaNtoso/akn:judgment/akn:meta/akn:identification/akn:FRBRWork/akn:FRBRdate,
-    <akn:FRBRdate date="{$content}" name="judgment" xmlns="http://docs.oasis-open.org/legaldocml/ns/akn/3.0"/>)
+    (: ...iterate over the FRBRdate element(s) :)
+    for $frbr_date in doc($uri)/akn:akomaNtoso/akn:judgment/akn:meta/akn:identification/akn:FRBRWork/akn:FRBRdate
+        (: Store the existing @name attribute value :)
+        let $attribute_name := $frbr_date/@name
+        (: Replace the exisitng FRBRdate element with a new one, using the same @name attribute :)
+        return xdmp:node-replace(
+        document($uri)/akn:akomaNtoso/akn:judgment/akn:meta/akn:identification/akn:FRBRWork/akn:FRBRdate[@name = $attribute_name],
```
dragon-dxw, line 17: can this be `$frbr_date` ?
Is there a difference between doc($uri) and document($uri) ?

dragon-dxw (commented):
I still don't like how we're setting a bunch of externally-meaningful dates to the same date, that feels wrong, but it's better than the current bug and I'm happy if Jim and you are happy.

PR #63, opened 2022-08-08
Respect & maintain the `FRBRdate` attribute `@name`

lozette:
## Changes in this PR:

Relevant ADR: https://github.com/nationalarchives/ds-caselaw-public-access-service/blob/main/doc/adr/0008-how-frbrdates-are-managed.md

Related to https://github.com/nationalarchives/ds-caselaw-public-ui/pull/316

This XQuery updates the nodes at `/akn:akomaNtoso/akn:judgment/akn:meta/akn:identification/akn:FRBRWork/akn:FRBRdate`
and `/akn:akomaNtoso/akn:judgment/akn:meta/akn:identification/akn:FRBRExpression/akn:FRBRdate`

The XQuery is used in the editor UI, and is called when the editors update a
judgment or tribunal's "date". This is the `akn:FRBRWork/akn:FRBRdate` date as
displayed in the public-ui search results.

As per the ADR and schema, `akn:FRBRWork/akn:FRBRdate` and `akn:FRBRExpression/akn:FRBRdate`
should be kept in sync, hence this XQuery updates both.

Previously, the Xquery did not respect the value of the `@name` attribute. It was being
set to `judgment`, which is true in many cases, but may also be `hearing` in
some cases.

In the referred PR above, we changed the way we get a judgment's date, from
looking for an `FRBRdate` with a name attribute of `judgment`, we instead
look for any `FRBRdate` inside the `FRBRWork` element. So the attribute name
value is redundant.

When updating the `FRBRdate` within `FRBRExpression and `FRBRWork`, we now:

1) Keep both dates in sync
2) Update the `name` attribute of both to match the value at `judgment/@name`.

This is because an `FRBRdate` element *may* have a value of "dummy" if the parser
failed to create a date for the judgment/tribunal. So we want to update it to
match the judgment/tribunal "type". This has been confirmed by Jim @jurisdatum

This change also assumes a judgment/tribunal ALWAYS has a `FRBRdate` element
within `FRBRWork` and `FRBRExpression`, as a missing date element is a parser
error.

https://trello.com/c/lTTnAqpz/847-date-editor-ui


dragon-dxw (approved):
Don't know if it'd be useful to put some sort of comment pointing at the ADR, or commenting on the fact that they're the same (but they might not want to be).

Congratulations on getting this nailed down.

dragon-dxw:
New changes look like an improvement. Someone's going to wonder why something that has nothing to do with a date is the date_attribute, but the commit message covers that admirably.

PR #67, opened 2022-08-11
Add methods to get the lock/checkout status of a judgment

lozette:
## Changes in this PR:

In the privileged API, we want to be able to see the checkout/lock status of a
judgment.

`get_judgment_checkout_status.xqy` checks the checkout status of a judgment using
`dls:document-checkout-status($uri)`

The result will either be empty, if the judgment is not locked, or an XML
fragment containing the URI and annotation message, if the judgment is locked.

The method `get_judgment_checkout_status` can be called on the Client, and will
return a response with either an XML fragment describing who has locked the
judgment, or empty if it is not locked.

The method `get_judgment_checkout_status_message` can also be called on the
Client, and will return either the lock annotation, if the judgment is locked,
or "not locked" if the judgment is not locked.

https://trello.com/c/l60TQkpe/261-implement-checkout-endpoint
https://trello.com/c/nXl9N4Iz/840-add-checkout-status-function-to-ml-client

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -333,6 +333,29 @@ class MarklogicApiClient:
             accept_header="application/xml",
         )
 
+    def get_judgment_checkout_status(self, judgment_uri: str) -> requests.Response:
+        uri = f"/{judgment_uri.lstrip('/')}.xml"
+        xquery_path = os.path.join(
+            ROOT_DIR, "xquery", "get_judgment_checkout_status.xqy"
+        )
+        vars = {
+            "uri": uri
+        }
+
+        return self.eval(
+            xquery_path,
+            vars=json.dumps(vars),
+            accept_header="application/xml",
+        )
+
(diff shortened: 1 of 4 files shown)


src/caselawclient/Client.py
```diff
@@ -331,6 +333,22 @@ def checkin_judgment(self, judgment_uri: str) -> requests.Response:
...
+        vars = {
+            "uri": uri
+        }
+
+        response = self.eval(
+            xquery_path,
+            vars=json.dumps(vars),
+            accept_header="application/xml",
+        )
+        return self.parse_checkout_status_response(response.text)
```
dragon-dxw, line 350: This function feels like an outlier, in that it doesn't return an `eval()`d response, but a parsed string.

There might be value in extracting a common "call remote XQuery" function but equally we might not be ready for that.

dragon-dxw (approved):

dragon-dxw (approved):

PR #69, opened 2022-08-14
Judgment checkout may optionally expire at midnight

lozette:
## Changes in this PR:

A comment on https://trello.com/c/l60TQkpe suggests the checkout/lock of
a judgment may expire at midnight, to ensure judgments aren't locked
"forever".

This commit adds a function `calculate_seconds_until_midnight` which calculates
how many seconds there are until midnight.

The Marklogic function `dls:document-checkout` contains an optional third
parameter which lets you pass in the amount of time in seconds a document
should be locked for.

Marrying these two up, we have amended `checkout_judgment` on the client to
take an optional `expires_at_midnight` parameter. Passing True to this param
means that the number of seconds until midnight is calculated, and passed to
`dls:document-checkout`.

If `expires_at_midnight` is not passed, an empty sequence `()` is used in
`dls:document-checkout` and the judgment is locked indefinitely.

 https://trello.com/c/l60TQkpe 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog 1.0.0].
 ## [Unreleased]
 - Add optional annotation parameter to `checkout_judgment` method
 - Add method to get the lock/checkout status of a judgment
+- Judgment checkout may optionally expire at midnight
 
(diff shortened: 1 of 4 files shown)


dragon-dxw (approved):
👍

PR #74, opened 2022-08-18
Add linting 

lozette:
## Changes in this PR:

Add linting to this repo.

Add some automatic fixes from the linters.

Change/remove some typing to appease MyPy.

Deprecate two methods, client.GET() and client.POST() as we do not believe they are being used. If they are not, they can be removed later (along with make_request())


requirements.txt
```diff
@@ -4,4 +4,5 @@ django-environ==0.8.1
 idna==3.3
 requests==2.28.1
 requests-toolbelt==0.9.1
+types-requests==2.28.9
```
dragon-dxw, line 7: We probably don't need this line now we've got it in the pre-commit (pre-commit handles their own dependency/install jazz); note the lack of mypy in this file

dragon-dxw (approved):

PR #75, opened 2022-08-18
Ensure only people who are allowed to view unpublished judgments can view them

lozette:
## Changes in this PR:

We have added `user_can_view_unpublished_judgments(self.username)` to the
client, which allows us to check that the currently logged in user (`username`)
has the appropriate permission to view unpublished judgments.

If a user attempts to access unpublished judgments by (somehow) calling an
endpoint with `show_unpublished=True` BUT they do not have the correct privilege
to view unpublished judgments, they should not be allowed to do that. Instead
`show_unpublished` will be set to False and they will not be able to see
unpublished judgments.

If a user who *is* allowed to view unpublished judgments asks specifically to
view unpublished judgments, by passing in `show_unpublished=False`, the value
of `show_unpublished` will remain False to allow them to view the required
unpublished judgments.

TODO: Refactor these tests!!

https://trello.com/c/o3Wbto2N/263-implement-preview-permissions

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -183,8 +183,7 @@ class MarklogicApiClient:
         self, judgment_uri, version_uri=None, show_unpublished=False
     ) -> str:
         uri = f"/{judgment_uri.lstrip('/')}.xml"
-        if not self.user_can_view_unpublished_judgments(self.username):
-            show_unpublished = False
+        show_unpublished = self.verify_show_unpublished(show_unpublished)
         if version_uri:
             version_uri = f"/{version_uri.lstrip('/')}.xml"
         xquery_path = os.path.join(ROOT_DIR, "xquery", "get_judgment.xqy")
@@ -502,10 +501,8 @@ class MarklogicApiClient:
         :param only_unpublished: If True, will only return published documents. Ignores the value of show_unpublished
         :return:
         """
-
         module = "/judgments/search/search.xqy"  # as stored on Marklogic
-        if not self.user_can_view_unpublished_judgments(self.username):
-            show_unpublished = False
+        show_unpublished = self.verify_show_unpublished(show_unpublished)
         vars = json.dumps(
             {
                 "court": str(court or ""),
@@ -523,7 +520,6 @@ class MarklogicApiClient:
                 "only_unpublished": str(only_unpublished).lower(),
             }
         )
-
         return self.invoke(module, vars)
 
     def eval_xslt(
@@ -542,8 +538,7 @@ class MarklogicApiClient:
         else:
             image_location = ""
 
-        if not self.user_can_view_unpublished_judgments(self.username):
-            show_unpublished = False
+        show_unpublished = self.verify_show_unpublished(show_unpublished)
 
         vars = json.dumps(
             {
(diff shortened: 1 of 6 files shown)


src/caselawclient/Client.py
```diff
@@ -150,6 +151,7 @@ def POST(
 
     def get_judgment_xml(self, judgment_uri, version_uri=None, show_unpublished=False) -> str:
         uri = f"/{judgment_uri.lstrip('/')}.xml"
+        show_unpublished = self.verify_show_unpublished(show_unpublished)
```
dragon-dxw, line 154: I find the logic here really complicated (especially the type conversions); I don't know whether a specifically string-emitting version of this function would be useful, rather than handling it in the vars.
A bit of me doesn't like the shadowing of the "please show unpublished" by "you should show unpublished, although not having that potentially misleading variable hanging around is good.

tests/test_client.py
```diff
@@ -33,79 +34,186 @@ def test_eval_calls_request(self, MockPath):
...
+                        "page-size": 20,
+                        "q": "my-query",
+                        "party": "a party",
+                        "neutral_citation": "",
+                        "specific_keyword": "",
+                        "order": "",
+                        "from": "",
+                        "to": "",
+                        "show_unpublished": "false",
+                        "only_unpublished": "false"
```
dragon-dxw, line 142: Does it make more sense to raise a (probably custom) error here than automatically downgrade? (I think that raising breaks things later on, perhaps?)

tests/test_client.py
```diff
@@ -33,79 +34,186 @@ def test_eval_calls_request(self, MockPath):
...
+                    }
+
+                    client.invoke.assert_called_with(
+                        '/judgments/search/search.xqy',
+                        json.dumps(expected_vars)
+                    )
+                    mock_logging.assert_called()
+
+    def test_eval_xslt_with_default_and_user_can_view_unpublished(self):
```
dragon-dxw, line 151: I'm not sure what 'default' is in this context

tests/test_client.py
```diff
@@ -571,4 +679,35 @@ def test_user_can_view_unpublished_judgments_false(self):
             client.eval.return_value.text = 'false'
 
             result = client.user_can_view_unpublished_judgments("laura")
+            self.assertEqual(result, False)
+
+    def test_verify_show_unpublished_user_cannot_view_unpublished_and_show_unpublished_true(self):
```
dragon-dxw, line 684: I've finally parsed these test names and I think it might be worth rewording them:

something like:
test_verifiy_forbids_unpublished_if_unauthorised_and_asks
test_verify_allows_unpublished_if_authorised_and_asks
test_verify_forbids_unpublished_if_unauthorised_and_doesnt_ask
test_verify_forbids_unpublished_if_authorised_and_doesnt_ask

dragon-dxw: Definately worth looking into how we do contexts for these :)

tests/test_client.py
dragon-dxw: Oh, it's "show_unpublished" being true, in the meaning of "please show unpublished (if I'm allowed)"

dragon-dxw (approved):
Probably worth a videochat next week, as this is tough and important to get right, but I'm pretty happy

dragon-dxw:
There's a TODO: refactor those tests in the description, which we should address and will be easy to overlook :)

tests/test_client.py
lozette: No, "default" is the default XSLT as opposed to the accessible one 🤦  I realise these are so confusing. I will look into contexts/better naming.

PR #77, opened 2022-08-24
Break judgment checkout

lozette:
## Changes in this PR:

Add the ability to break a checkout lock on a judgment. If a judgment is locked
we need a way to break the lock and return it to an editable state.

The XQuery uses `dls:break-checkout($uri, $deep)` to break the lock and
any locks on associated versions.

The function `break_checkout()` returns a request object containing no content
- the XQuery returns an empty sequence (assuming the requested URI exists).

https://trello.com/c/3eDUGYfk/837-implement-break-checkout-endpoint-function

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -694,6 +694,20 @@ class MarklogicApiClient:
             accept_header="application/xml",
         )
 
+    def break_checkout(self, judgment_uri):
+        uri = self._format_uri(judgment_uri)
+        xquery_path = os.path.join(ROOT_DIR, "xquery", "break_judgment_checkout.xqy")
+        vars = json.dumps(
+            {
+                "uri": uri,
+            }
+        )
+        return self.eval(
+            xquery_path,
+            vars=vars,
+            accept_header="application/xml",
+        )
+
     def user_has_privilege(self, username, privilege_uri, privilege_action):
         xquery_path = os.path.join(ROOT_DIR, "xquery", "user_has_privilege.xqy")
         vars = json.dumps(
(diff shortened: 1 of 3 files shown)


src/caselawclient/xquery/break_judgment_checkout.xqy
```diff
@@ -0,0 +1,7 @@
+xquery version "1.0-ml";
+import module namespace dls = "http://marklogic.com/xdmp/dls" at "/MarkLogic/dls.xqy";
+
+declare variable $uri as xs:string external;
+let $deep := fn:true()
+
+return dls:break-checkout($uri, $deep)
```
dragon-dxw, line 7: > deep
>  If set to true, then the checkout is broken for any checked-out documents that are directly or indirectly included by the specified document.

Seems like a potentially wide-ranging unlock by default, but I don't think we have included documents yet, and it'd be a pain if something wasn't unlocked when we wanted it to be.

src/caselawclient/Client.py
```diff
@@ -694,6 +694,20 @@ def copy_judgment(self, old, new):
             accept_header="application/xml",
         )
 
+    def break_checkout(self, judgment_uri):
+        uri = self._format_uri(judgment_uri)
+        xquery_path = os.path.join(ROOT_DIR, "xquery", "break_judgment_checkout.xqy")
+        vars = json.dumps(
+            {
+                "uri": uri,
```
dragon-dxw, line 702: we may want to come back and allow 'deep' to be set in Python, but no need to add it now.

dragon-dxw (approved):

PR #78, opened 2022-08-25
Refactor tests

lozette:
## Changes in this PR:

`test_client` had reached over 700 lines. Refactor it for clarity.

1) Split the tests for the client and xml_tools into their own folders.
2) Split the tests in `test_client` into multiple files. These are divided into groups according to the types of functions they test: getting judgments, getting/setting properties, checking judgments in & out etc
3) Reduce the number of repeated assertions on several tests, for brevity.


dragon-dxw (approved):
All looks very sensible. Thank you!

PR #79, opened 2022-08-26
Support renaming of the XSL Transformation files

lozette:
## Changes in this PR:

In https://github.com/nationalarchives/ds-caselaw-public-access-service/pull/72
we will be renaming the XSL Transformation files in Marklogic from `judgment0`
and `judgment2` to `as-handed-down` and `accessible-html`, in order to be more
descriptive. This commit supports those changes.

https://trello.com/c/dylwn8lt/830-rename-judgment2-and-judgment0-xslts-in-public-access-service


lozette:
The renamed XSLTs are now in Marklogic on prod & staging 
Screenshot 2022-09-01 at 11 49 29

dragon-dxw (approved):

PR #81, opened 2022-09-05
DRY up some aspects of the API Client

lozette:
## Changes in this PR:

`Client.py` is getting long. We should look into splitting/reducing it.

In the meantime, use three methods to DRY up the code a little:

1. We already had `self._format_uri()` but weren't using it everywhere. Use it everywhere!
2. We were repeatedly using `os.path.join` to get the XQuery file path. Add `self._xquery_path()` to remove this duplicated code and remove a few lines. 
3. Standardise all calls to `eval()` and then bundle them up into a helper method `self._send_to_eval()` and use this on all method calls.

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -138,7 +138,11 @@ class MarklogicApiClient:
             raise new_exception
 
     def _format_uri(self, uri):
-        return f"/{uri.lstrip('/')}.xml"
+        """
+        Marklogic requiores a document URI that begins with a slash `/` and ends in `.xml`.
+        This method ensures any URI passed into the client matches this format.
+        """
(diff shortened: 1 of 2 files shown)


src/caselawclient/Client.py
```diff
@@ -182,22 +192,17 @@ def POST(
     def get_judgment_xml(
         self, judgment_uri, version_uri=None, show_unpublished=False
     ) -> str:
-        uri = f"/{judgment_uri.lstrip('/')}.xml"
+        uri = self._format_uri(judgment_uri)
```
dragon-dxw, line 195: Not one to fix now, necessarily, but I'm struck by how `_format_uri` says nothing about the target format. Should probably contain `xml` in the name somewhere

dragon-dxw (approved):

PR #82, opened 2022-09-07
Release v4.10.0

lozette:
Making this one a minor release as even though there are no changes the end users will see, there's been a lot of re-jigging under the hood.

- Add linting
- Ensure only people who are allowed to view unpublished judgments can view them
- Refactor tests
- Break judgment checkout
- Methods & XQueries to get & set all metadata
- DRY up some aspects of the API Client
- Support renaming of the XSL Transformation files
- Add memoization to speed up privilege checking

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog 1.0.0].
 
 ## [Unreleased]
+
+## [Release 4.10.0]
+- Add linting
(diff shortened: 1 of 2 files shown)


dragon-dxw:
I think I'd very much like at least the 
[Don't check privilege if user does not ask for unpublished judgments](https://github.com/nationalarchives/ds-caselaw-custom-api-client/pull/84/commits/78d652bc5dd7d36712a4062e3ea1e85c9a85628b) commit in this to avoid significant performance regressions on public-ui live.

PR #84, opened 2022-09-07
Speed up privilege checking

lozette:
## Changes in this PR:

When testing the API Client before a release, we noticed that the performance of the editor UI had degraded. The culprit is `verify_show_unpublished()` which checks that the active user has the right to view unpublished judgments. This is being called every time a user tried to view a judgment, or lists judgments. As a result, it was being called 10 times for each search result page. 

Our quick and inelegant (and hopefully temporary) solution is to use a rudimentary cache, which will check to see if the user has been through `user_can_view_unpublished_judgments` already, and if so use the result of that which is stored in the cache. 

This has made the editor UI faster in local testing.

The ultimate aim is to replace this with privilege checking within Marklogic.

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -10,6 +10,7 @@ from xml.etree.ElementTree import Element
 
 import environ
 import requests
+from memoization import cached
 from requests.auth import HTTPBasicAuth
 from requests_toolbelt.multipart import decoder
 
@@ -19,8 +20,6 @@ env = environ.Env()
 RESULTS_PER_PAGE = 10
 ROOT_DIR = os.path.dirname(os.path.realpath(__file__))
 DEFAULT_XSL_TRANSFORM = "accessible-html.xsl"
-
-user_read_unpublished_cache: Dict[str, bool] = {}
 
 
 class MarklogicAPIError(requests.HTTPError):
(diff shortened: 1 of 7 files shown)


src/caselawclient/Client.py
```diff
@@ -625,6 +627,14 @@ def user_can_view_unpublished_judgments(self, username):
         result = multipart_data.parts[0].text
         return result.lower() == "true"
 
+    def user_can_view_unpublished_judgments_cached(self, username):
+        print(user_read_unpublished_cache)
```
Floppy, line 631:
```suggestion
```

src/caselawclient/Client.py
```diff
@@ -625,6 +627,14 @@ def user_can_view_unpublished_judgments(self, username):
         result = multipart_data.parts[0].text
         return result.lower() == "true"
 
+    def user_can_view_unpublished_judgments_cached(self, username):
```
Floppy, line 630: Personally I wouldn't have a separate `_cached` implementation - probably I would have change `user_can_view_unpublished_judgements` to a private method and rename this one to that name. External users of the class don't need to care about the uncached version. But that's just personal OO style, it's certainly not a *necessary* change.

Floppy (changes requested):

Floppy:
As an alternative option, in Ruby I'd probably use something like [Memoist](https://github.com/matthewrudy/memoist) for all the boilerplate of caching results. Looks like https://pypi.org/project/memoization/ is a python version.

But then I'm a big fan of pulling in useful dependencies, you may well prefer the explicit approach you've taken here so as not to confuse the reader with works like "memoization". I don't mind. :)

Floppy (approved):

PR #86, opened 2022-09-21
Page 0 redirects to page 1

dragon-dxw:
If a non-positive page is selected, force it to be the first page instead.


lozette (approved):
There's a card for this one https://trello.com/c/T3Lg06mJ/943-searching-for-page-0-in-the-atom-feed-generates-an-error

PR #90, opened 2022-10-03
Admin users can't read unpublished judgments

lozette:
## Changes in this PR:

Allow users who have the `admin` role to view unpublished judgments.
This feels slightly backwards to me but unfortunately the `admin` role is not
editable in Marklogic, so granting the `admin` user the correct custom privilege
to view unpublished judgments is not immediately obvious.

As we have now figured out our custom roles in a previous PR (
https://github.com/nationalarchives/ds-caselaw-public-access-service/pull/86)
this work seems less urgent (as we can tailor our custom roles and should not
need to grant admin access to users). But it is still useful to be able to view
unpublished judgments as an admin user, for completeness.

There may be a way to grant this access via Marklogic's QBAC system
https://docs.marklogic.com/guide/security/qbac - for future investigation.

https://trello.com/c/Z7IXPWVq

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -637,6 +637,9 @@ class MarklogicApiClient:
 
     @cached
     def user_can_view_unpublished_judgments(self, username):
+        if self.user_has_admin_role(username):
+            return True
+
         check_privilege = self.user_has_privilege(
             username,
             "https://caselaw.nationalarchives.gov.uk/custom/privileges/can-view-unpublished-documents",
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -646,6 +646,23 @@ class MarklogicApiClient:
         result = multipart_data.parts[0].text
         return result.lower() == "true"
 
+    def user_has_role(self, username, role):
+        vars = {
+            "user": username,
+            "role": role,
+        }
+        return self._send_to_eval(vars, "user_has_role.xqy")
+
+    @cached
+    def user_has_admin_role(self, username):
+        check_role = self.user_has_role(
+            username,
+            "admin",
+        )
+        multipart_data = decoder.MultipartDecoder.from_response(check_role)
+        result = multipart_data.parts[0].text
+        return result.lower() == "true"
+
     def calculate_seconds_until_midnight(self, now=None):
         """
         Get timedelta until end of day on the datetime passed, or current time.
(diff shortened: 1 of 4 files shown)


src/caselawclient/Client.py
```diff
@@ -637,6 +637,9 @@ def user_has_privilege(self, username, privilege_uri, privilege_action):
 
     @cached
     def user_can_view_unpublished_judgments(self, username):
+        if self.user_has_admin_role(username):
```
dragon-dxw, line 640: 1. Does production public ui now use a non-admin user? Wary about whether this will let people potentially see unpublished documents. (Should be fine but not 100% convinced)

2. Was concerned about multiple requests to work out if a user can do a thing, but it's cached so that should be fine.

src/caselawclient/Client.py
```diff
@@ -646,6 +649,23 @@ def user_can_view_unpublished_judgments(self, username):
...
         return result.lower() == "true"
 
+    def user_has_role(self, username, role):
+        vars = {
+            "user": username,
+            "role": role,
+        }
+        return self._send_to_eval(vars, "user_has_role.xqy")
+
+    @cached
```
dragon-dxw, line 659: Does it make more sense to cache user_has_role instead?

src/caselawclient/Client.py
```diff
@@ -646,6 +649,23 @@ def user_can_view_unpublished_judgments(self, username):
...
+    def user_has_admin_role(self, username):
```
dragon-dxw, line 660: I'd be tempted to reword this so that the decoder parts are inside `user_has_role` which can then return a clean True/False answer

dragon-dxw (commented):
I like it, a few comments I'd like answers for before approval

src/caselawclient/Client.py
lozette: No, production doesn't yet have the apps using their custom users, I was going to wait until after office hours to move the apps over.

src/caselawclient/Client.py
lozette: I'm in two minds about this, as some of the methods in the client return a response and some return strings, and there's no rhyme or reason about which ones do which. Most seem to return a response which is then parsed by something else.

src/caselawclient/Client.py
lozette: I don't know - `user_has_privilege` is not cached while `user_can_view_unpublished_judgments` is, so I was mirroring that. Can you let me know your thoughts on the privilege based ones? I think if we're going to be caching  `user_has_role` instead of its implementing method, then we should do it for `user_has_privilege` as well?

dragon-dxw: The mirroring is good and it's worth keeping. I think it does make sense to cache the True/False response, rather than the larger response we get back.

dragon-dxw (approved):

PR #92, opened 2022-10-04
Fix `TypeError: 'type' object is not subscriptable`

lozette:
I was getting `TypeError: 'type' object is not subscriptable` when running the tests and this seems to be the fix.


dragon-dxw (approved):
Sorry about that!

PR #94, opened 2022-10-13
Ensure Citations output as text rather than raw MarkLogic xml

dragon-dxw:
The get_judgment_citation function doesn't seem to be used anywhere else, so we can just change its output.

### DIFF
--- a/tests/client/test_get_set_metadata.py
+++ b/tests/client/test_get_set_metadata.py
@@ -7,22 +7,23 @@ from unittest.mock import patch
 from src.caselawclient.Client import ROOT_DIR, MarklogicApiClient
 
 
+@patch("src.caselawclient.Client.decode_multipart")
+@patch("src.caselawclient.Client.MarklogicApiClient.eval")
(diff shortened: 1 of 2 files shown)


dragon-dxw:
[BUG in API client on querying neutral citation for a single judgment](https://trello.com/c/BlZJAZ1o/1076-bug-in-api-client-on-querying-neutral-citation-for-a-single-judgment)

tests/client/test_get_set_metadata.py
```diff
@@ -7,23 +7,24 @@
 from src.caselawclient.Client import ROOT_DIR, MarklogicApiClient
 
 
+@patch("src.caselawclient.Client.MarklogicApiClient.eval")
+def test_get_judgment_citation(send):
+    uri = "judgment/uri"
+    expected_vars = {"uri": "/judgment/uri.xml"}
+    send.return_value = 6
```
lozette: Can this return value be something more citation-esque?

lozette (approved):

PR #98, opened 2022-11-08
WIP: Catch AttributeError if response is none

lozette:

https://rollbar.com/dxw/tna-caselaw-editor-ui/items/70/?utm_campaign=exp_repeat_item_message&utm_medium=slack&utm_source=rollbar-notification


dragon-dxw (approved):
Why does the decode fail? Is it because parts is an empty list?

We could say

```
if part_count>0:
   return multipart_data.parts[0].text
else:
   return None
```
   
but I'm not sure that's substantially cleaner.

lozette:
I don't know I'm afraid, this was a 10 minute hypothesis!

PR #113, opened 2023-01-09
Expose a MarklogicValidationFailed exception to support enrichment

dragon-dxw:
We expose a new exception, `MarklogicValidationFailedError` which corresponds to a class of Marklogic errors starting `XDMP-VALIDATE`...

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import os
+import re
 import warnings
 from datetime import datetime, time, timedelta
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


src/caselawclient/Client.py
```diff
@@ -213,7 +228,7 @@ def make_request(
         data: Optional[Dict[str, Any]] = None,
     ) -> requests.Response:
         kwargs = self.prepare_request_kwargs(method, path, body, data)
-        self.session.headers = headers  # type: ignore
```
dragon-dxw, line 216: For some reason, this annotation is no longer valid in mypy on Github, but is mandatory locally.

jacksonj04 (approved):

PR #116, opened 2023-01-16
Check content hash

dragon-dxw:
Implements https://github.com/nationalarchives/tna-judgments-parser/blob/main/src/akn/SHA256.cs

Gets the text of the judgment (ignoring the metadata section) and strips out all whitespace and SHA256 hashes it.

Had some problems with getting the linters to play nice: see #118 fora resolution.

* Is this the right place in the code base?
* Need to wire this up to stuff -- is that best done here or in priv-api etc.

### DIFF
--- a/tests/content_hash/test_content_hash.py
+++ b/tests/content_hash/test_content_hash.py
@@ -1,5 +1,8 @@
-from src.caselawclient.content_hash import (content_hash, hashable_text,
-                                            validate_content_hash)
+from src.caselawclient.content_hash import (
+    content_hash,
+    hashable_text,
+    validate_content_hash,
+)
 
 VALID_DOC = b"""<?xml version="1.0" encoding="UTF-8"?>
     <akomaNtoso
--- a/tests/content_hash/test_content_hash.py
+++ b/tests/content_hash/test_content_hash.py
@@ -1,0 +1,51 @@
+from src.caselawclient.content_hash import (content_hash, hashable_text,
+                                            validate_content_hash)
+
+VALID_DOC = b"""<?xml version="1.0" encoding="UTF-8"?>
+    <akomaNtoso
+      xmlns="http://docs.oasis-open.org/legaldocml/ns/akn/3.0"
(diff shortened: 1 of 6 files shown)


dragon-dxw:
[Validate content hash](https://trello.com/c/I7D7vtxC/274-validate-content-hash)

tests/content_hash/test_content_hash.py
```diff
@@ -0,0 +1,46 @@
+from src.caselawclient.content_hash import (content_hash, hashable_text,
```
anthonyhashemi, line 1: [formatting] needs isort/black

tests/content_hash/test_content_hash.py
```diff
@@ -0,0 +1,46 @@
...
+      </judgment>
+    </akomaNtoso>
+    """
+
+INVALID_DOC = VALID_DOC.replace(b"Do", b"Do not").replace(b"valid", b"invalid")
+
+
+def test_hashable_text():
+    """Do we correctly identify the text to hash, omitting the meta section and removing spaces?"""
+    assert hashable_text(VALID_DOC) == b"Dousethisvalidtext"
```
anthonyhashemi, line 26: [non-blocking] I know the two cases (note my suggestion of removing the third) here are pretty simple, but I feel it could be worth splitting out the valid, invalid test cases into their own separate test functions, just so that if one of these test cases starts failing in the future we will know whether the positive or negative case has failed.

And similarly for the other tests

Not a blocker though

tests/content_hash/test_content_hash.py
```diff
@@ -0,0 +1,46 @@
...
+    assert hashable_text(INVALID_DOC) == b"Donotusethisinvalidtext"
+    assert b"BAD" not in hashable_text(VALID_DOC)
```
anthonyhashemi, line 28: I think this 3rd assert can be removed given we already assert what the expected full string of the valid hashable_text is, so don't need to test if BAD is not in it

anthonyhashemi (approved):

PR #117, opened 2023-01-16
Dummy date set/read

dragon-dxw:
Is this the correct approach?
* An empty input string is saved as `<FRBRdate date='1000-01-01' name='dummy'>`
* MarkLogic returns`dummy` dates as `()`
* The Multipart Decoder returns the empty string if no parts are encountered (existing behaviour)

This feels slightly risky and might not be the right behaviour.

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -292,6 +292,7 @@ class MarklogicApiClient:
         return self.set_judgment_work_expression_date(judgment_uri, content)
 
     def set_judgment_work_expression_date(self, judgment_uri, content):
+        """Sets the date on both the Work and the Expression of the judgment to the given value"""
         uri = self._format_uri_for_marklogic(judgment_uri)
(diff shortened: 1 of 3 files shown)


dragon-dxw:
[Ensure editor-ui inserts valid date](https://trello.com/c/9uIJUkB8/314-ensure-editor-ui-inserts-valid-date)

Floppy:
This seems OK, the logic in the code is sound given the approach. The only other approach I can think of is to either leave out the FRBRDate tag (which I think isn't allowed in the schema), or to leave out the attributes (but I can't tell if that's allowed or not). I think if it IS valid, having no attributes is probably the best approach, but if that's not allowed then this is as good as we can get.

anthonyhashemi:
@dragon-dxw  I have a brief memory about this but it's been a while, do you know if this change is it still required? I see Heather archived the card a while back but potentially still a problem?

PR #122, opened 2023-02-01
Add content hash validation when we save a locked judgment

dragon-dxw:
(i.e. when we receive something from the privileged API)

Includes pulling errors into another file to avoid circular references

### DIFF
--- a/tests/client/test_save_copy_delete_judgment.py
+++ b/tests/client/test_save_copy_delete_judgment.py
@@ -35,6 +35,11 @@ class TestSaveCopyDeleteJudgment(unittest.TestCase):
             )
 
     def test_save_locked_judgment_xml(self):
+        """
+        Given a locked judgement uri, a judgement_xml and an annotation
+        When `Client.save_locked_judgment_xml` is called with these as arguments
+        Then the xquery in `update_locked_judgment.xqy` is called on the Marklogic db with those arguments
+        """
         with patch.object(src.caselawclient.Client, "validate_content_hash"):
             with patch.object(self.client, "eval"):
                 uri = "/ewca/civ/2004/632"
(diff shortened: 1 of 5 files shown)


tests/content_hash/test_content_hash.py
```diff
@@ -24,31 +27,42 @@
...
+        assert hashable_text(VALID_DOC) == b"Dousethisvalidtext"
+
+    def test_hashable_text_invalid_doc(self):
+        assert hashable_text(INVALID_DOC) == b"Donotusethisinvalidtext"
 
 
-def test_hashable_text_invalid_doc():
-    assert hashable_text(INVALID_DOC) == b"Donotusethisinvalidtext"
+class TestCorrectHashForString:
+    def test_content_hash(self):
```
anthonyhashemi, line 43: [minor]: could split out cases for clarity

tests/client/test_save_copy_delete_judgment.py
```diff
@@ -31,23 +35,35 @@ def test_save_judgment_xml(self):
             )
 
     def test_save_locked_judgment_xml(self):
-        with patch.object(self.client, "eval"):
+        with patch.object(src.caselawclient.Client, "validate_content_hash"):
+            with patch.object(self.client, "eval"):
```
anthonyhashemi, line 39: could consider adding docstrings for these tests.

For example for this test:

```
"""
Given a locked judgement uri, a judgement_xml and an annotation
When `Client.save_locked_judgment_xml` is called with these as arguments
Then the xquery in `update_locked_judgment.xqy` is called on the Marklogic db with those arguments
"""
```

anthonyhashemi (approved):

PR #124, opened 2023-03-09
Do not default to current version if version doesn't exist

dragon-dxw:
[Trello](https://trello.com/c/VIVTctiL/534-make-invalid-version-uris-return-an-error-state)

`fn:doc` of a non-existant node returns `()`, which looks the same as no version doc being passed in the first place

### DIFF
--- a/src/caselawclient/xquery/xslt_transform.xqy
+++ b/src/caselawclient/xquery/xslt_transform.xqy
@@ -6,21 +6,13 @@ declare variable $version_uri as xs:string? external;
 declare variable $img_location as xs:string? external;
 declare variable $xsl_filename as xs:string? external;
 
-let $judgment_xml := fn:doc($uri)
-let $version_xml := if ($version_uri) then fn:document($version_uri) else ()
 let $judgment_published_property := xdmp:document-get-properties($uri, xs:QName("published"))[1]
 let $is_published := $judgment_published_property/text()
 let $image_base := if ($img_location) then $img_location else ""
-let $xsl := if ($xsl_filename) then $xsl_filename else "judgment2.xsl"
-
-let $document_to_transform := if ($version_xml) then $version_xml else $judgment_xml
-let $xsl_path := fn:concat("judgments/xslts/", $xsl)
+let $document_to_transform := if ($version_uri) then fn:document($version_uri) else fn:doc($uri)
+let $xsl_path := fn:concat("judgments/xslts/", $xsl_filename)
 
 let $params := map:map()
-let $_put := map:put(
-                    $params,
-                    "image-base",
-                    $image_base)
 
 let $return_value := if (xs:boolean($is_published) or $show_unpublished) then
         xdmp:xslt-invoke($xsl_path,
--- a/src/caselawclient/xquery/xslt_transform.xqy
+++ b/src/caselawclient/xquery/xslt_transform.xqy
@@ -14,6 +14,11 @@ let $xsl_path := fn:concat("judgments/xslts/", $xsl_filename)
 
 let $params := map:map()
 
+let $_ := if (not(exists($document_to_transform))) then
+  (
+    fn:error(xs:QName("FCL_DOCUMENTNOTFOUND"), "No XML document was found to transform")
+  ) else ()
+
 let $return_value := if (xs:boolean($is_published) or $show_unpublished) then
         xdmp:xslt-invoke($xsl_path,
           $document_to_transform,


src/caselawclient/xquery/xslt_transform.xqy
```diff
@@ -6,21 +6,13 @@ declare variable $version_uri as xs:string? external;
 declare variable $img_location as xs:string? external;
 declare variable $xsl_filename as xs:string? external;
 
-let $judgment_xml := fn:doc($uri)
-let $version_xml := if ($version_uri) then fn:document($version_uri) else ()
 let $judgment_published_property := xdmp:document-get-properties($uri, xs:QName("published"))[1]
 let $is_published := $judgment_published_property/text()
 let $image_base := if ($img_location) then $img_location else ""
-let $xsl := if ($xsl_filename) then $xsl_filename else "judgment2.xsl"
```
dragon-dxw, line 14: judgment2.xsl does not exist anywhere within the FCL project.

src/caselawclient/xquery/xslt_transform.xqy
```diff
@@ -6,21 +6,13 @@ declare variable $version_uri as xs:string? external;
...
-
-let $document_to_transform := if ($version_xml) then $version_xml else $judgment_xml
-let $xsl_path := fn:concat("judgments/xslts/", $xsl)
+let $document_to_transform := if ($version_uri) then fn:document($version_uri) else fn:doc($uri)
+let $xsl_path := fn:concat("judgments/xslts/", $xsl_filename)
 
 let $params := map:map()
-let $_put := map:put(
```
dragon-dxw, line 20: we never use _put; it feels like noise.

dragon-dxw:
Possible improvements:
* if not exists(doc($uri)) , fn:error('FCLNOSUCHVERSION')
* catch that or the existing error and 404 not 500

dragon-dxw:
[Make invalid version URIs return an error state](https://trello.com/c/VIVTctiL/534-make-invalid-version-uris-return-an-error-state)

dragon-dxw:
Needs refactoring to use the version number, not the version URI; comparing the URIs is not trivial.

src/caselawclient/xquery/xslt_transform.xqy
```diff
@@ -6,21 +6,18 @@ declare variable $version_uri as xs:string? external;
...
-                    $params,
-                    "image-base",
-                    $image_base)
+
+let $_ := if (not(exists($document_to_transform))) then
```
anthonyhashemi, line 17: do we have to define this redundant _ variable because xquery only allows conditional when defining variables?

src/caselawclient/xquery/xslt_transform.xqy
```diff
@@ -6,21 +6,18 @@ declare variable $version_uri as xs:string? external;
...
+  (
+    fn:error(xs:QName("FCL_DOCUMENTNOTFOUND"), "No XML document was found to transform")
```
anthonyhashemi, line 19: so this is where we explicitly raise the error we wanted in this situation

src/caselawclient/xquery/xslt_transform.xqy
```diff
@@ -6,21 +6,18 @@ declare variable $version_uri as xs:string? external;
...
-let $judgment_xml := fn:doc($uri)
-let $version_xml := if ($version_uri) then fn:document($version_uri) else ()
 let $judgment_published_property := xdmp:document-get-properties($uri, xs:QName("published"))[1]
 let $is_published := $judgment_published_property/text()
 let $image_base := if ($img_location) then $img_location else ""
-let $xsl := if ($xsl_filename) then $xsl_filename else "judgment2.xsl"
-
-let $document_to_transform := if ($version_xml) then $version_xml else $judgment_xml
-let $xsl_path := fn:concat("judgments/xslts/", $xsl)
+let $document_to_transform := if ($version_uri) then fn:document($version_uri) else fn:doc($uri)
```
anthonyhashemi, line 12: the change here and removal of line
```
let $version_xml := if ($version_uri) then fn:document($version_uri) else ()
```

is just to make the code more concise right?

dragon-dxw: yep

src/caselawclient/xquery/xslt_transform.xqy
dragon-dxw: yep

src/caselawclient/xquery/xslt_transform.xqy
dragon-dxw: I believe so, yes. Everything returns a value, and that value needs to be consumed; whilst one could write (as a whole program)

`if (1) then (2) else (3)`

the return value would be consumed and returned to the caller.

I think.

anthonyhashemi (approved):
All good ✅

PR #128, opened 2023-04-05
Add Judgment object to API Client

jacksonj04:
EUI has a concept of the `Judgment` model, which gives a single interface for reading and (some) manipulation of a Judgment object.

This should really belong in the API, so that PUI and EUI both interact in the same way.


dragon-dxw (approved):

jacksonj04:
Explicitly acknowledging the low test coverage; the actual coverage of the methods is the same as in the EUI, it's just less than it should be.

Once EUI and PUI are using this code it will actually be a net increase for coverage, since we're only maintaining the functionality in one place.

PR #129, opened 2023-04-05
Add PEP-561 typing declaration

jacksonj04:
The existence of the `py.typed` file declares that this package implements PEP-561-compliant type annotations and is safe for type checkers (such as mypy) to evaluate.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,6 +23,7 @@ install_requires =
     memoization
 python_requires = >=3.8
 include_package_data = True
+zip_safe = False
 
 [options.packages.find]
 where = src


setup.cfg
```diff
@@ -23,6 +23,7 @@ install_requires =
     memoization
 python_requires = >=3.8
 include_package_data = True
+zip_safe = False
```
dragon-dxw, line 26: Is this line necessary? Apparently this is deprecated and we're probably OK with this being true since we don't have C plugins?

dragon-dxw: > it's a recommendation from mypy, because it won't unzip a zipped dependency to check the typing

dragon-dxw (approved):

setup.cfg
dragon-dxw: https://mypy.readthedocs.io/en/stable/installed_packages.html

PR #133, opened 2023-04-06
Move to Poetry for dependencies and build

jacksonj04:
Poetry presents a cleaner system for dependency management (and locking), creation and management of local environments with appropriate Python versions, and building for pypi.


pyproject.toml
```diff
@@ -1,3 +1,29 @@
...
+name = "ds-caselaw-custom-api-client"
+version = "5.2.6"
+description = "An API client for interacting with the Caselaw Marklogic instance"
+authors = ["The National Archives"]
+homepage = "https://github.com/nationalarchives/ds-caselaw-custom-api-client"
+keywords = ["national archives", "caselaw"]
+readme = "README.md"
+
+[tool.poetry.dependencies]
+python = "^3.9"
```
anthonyhashemi, line 11: do we have to remove an equivalent requirements.txt or similar?

anthonyhashemi (approved):

PR #157, opened 2023-04-13
Correct relative import in judgment model

jacksonj04:
This now correctly imports the judgment model, so that it doesn't run into a missing module problem when deployed.

### DIFF
--- a/src/caselawclient/models/judgments.py
+++ b/src/caselawclient/models/judgments.py
@@ -3,7 +3,7 @@ from functools import cached_property
 
 from requests_toolbelt.multipart import decoder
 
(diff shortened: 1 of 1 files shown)


src/caselawclient/models/judgments.py
```diff
@@ -3,8 +3,7 @@
 
 from requests_toolbelt.multipart import decoder
 
-from src.caselawclient.Client import MarklogicApiClient
-
+from ..Client import MarklogicApiClient
```
anthonyhashemi, line 6: If possible, I think it might be preferable to have the absolute import in case even if its just the ability to move this file without having to change the import in the future

would `from caselawclient.Client import MarklogicApiClient` work?

PR #163, opened 2023-04-17
Add strict typing for dicts used in MarkLogic queries

jacksonj04:
This introduces a new script, `build_xquery_type_dicts`, which looks through the `.xqy` files in `src/caselawclient/xquery` and uses their contents to generate a new `xquery_type_dicts.py`. This file strictly declares all required and optional parameters for each query as `TypedDict`s, which in this case are subclassed via `MarkLogicAPIDict`.

`Client.py` requires that all `dict`s passed to `_send_to_eval` are instances of `MarkLogicAPIDict`, which means they will fail if they are not one of the `TypedDict`s generated by the script.

As a final belt-and-braces, generating these typings is added to pre-commit; changing the parameters in a `.xqy` file will cause changes to be made in `xquery_type_dicts.py`, which will in turn cause mypy to fail if they are incompatible with the current implementation.

### DIFF
--- a/script/build_xquery_type_dicts
+++ b/script/build_xquery_type_dicts
@@ -1,0 +1,70 @@
+#!/usr/bin/python3
+
+import re
+from os import listdir
+from os.path import join
+
+DICTS_FILE_PREAMBLE = """# This file is built automatically with script/build_xquery_type_dicts.
+# DO NOT CHANGE IT MANUALLY!
+
+from typing import Any, Optional, TypedDict
+
+
+class MarkLogicAPIDict(TypedDict):
+    pass"""
+
+XQY_FILES_PATCH = join("src", "caselawclient", "xquery")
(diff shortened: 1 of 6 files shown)


script/build_xquery_type_dicts
```diff
@@ -0,0 +1,65 @@
...
+xquery_files = [f for f in listdir(XQY_FILES_PATCH) if f.endswith(".xqy")]
+
+with open(DICTS_FILES_PATCH, "w") as dicts_file:
+    dicts_file.write(DICTS_FILE_PREAMBLE)
+
+    for xquery_file in sorted(xquery_files):
+        with open(join(XQY_FILES_PATCH, xquery_file)) as file:
+            data = file.read()
+
+            matches = re.findall(r"declare variable \$(.+) as (.+) external;", data)
```
dragon-dxw, line 48: this feels potentially brittle; particularly whitespace changes and capitalisation

dragon-dxw (approved):

PR #169, opened 2023-04-18
Prevent publication in more invalid states

jacksonj04:
At the moment the only thing which prevents a publication is the presence of a hold flag, but this is not the only state in which publishing a document is an invalid operation.

Add some more condition checks to the publish flow around a document's name, court code and NCN.

## What next?

We should be able to expose a list of reasons a document isn't publishable, so we can relay them to editors.

### DIFF
--- a/tests/models/test_judgments.py
+++ b/tests/models/test_judgments.py
@@ -193,17 +193,34 @@ class TestJudgment:
         published_judgment.is_published = True
         assert published_judgment.status == JUDGMENT_STATUS_PUBLISHED
 
+    def test_has_name(self, mock_api_client):
+        judgment_with_name = Judgment("test/1234", mock_api_client)
+        judgment_with_name.name = "Judgment v Judgement"
+
+        judgment_without_name = Judgment("test/1234", mock_api_client)
+        judgment_without_name.name = ""
+
+        assert judgment_with_name.has_name is True
+        assert judgment_without_name.has_name is False
+
 
 class TestJudgmentPublication:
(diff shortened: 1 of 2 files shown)


tests/models/test_judgments.py
```diff
@@ -193,17 +193,103 @@ def test_judgment_status(self, mock_api_client):
...
-    def test_judgment_is_publishable_if_held(self, mock_api_client):
+    def test_judgment_not_publishable_if_held(self, mock_api_client):
         judgment = Judgment("test/1234", mock_api_client)
         judgment.is_held = True
 
         assert judgment.is_publishable is False
 
-    def test_judgment_is_publishable_if_not_held(self, mock_api_client):
+    def test_judgment_not_publishable_if_missing_name(self, mock_api_client):
+        judgment = Judgment("test/1234", mock_api_client)
```
dragon-dxw, line 262: I'd vaguely like a test that the default test Judgment fake, no modifications, is publishable. Many of these tests could pass but miss the point if that isn't true. (The explicit "everything set correctly, good to go" is great, don't change that)

dragon-dxw (approved):

jacksonj04:
I've done some poking around with having a default `Judgment` which passes, and because of the way the class and API are bolted together it's a massive pain to mock the right things.

There's another planned refactor when this model supports editing which will tidy this up and reduce a lot of boilerplate, so I'm inclined to leave this case out of the current tests (because it's already a well-tested class) unless we actually find a bug involving it.

PR #180, opened 2023-04-27
Allow dependencies to be "at least this version"

dragon-dxw:
Public UI was pinned to a specific version of requests because the version of requests was explicitly set in the custom-api-client; changing to allow non-major upgrades, since libraries should not be too picky about versions.

`poetry upgrade`d, to check things work with the most recent version.


jacksonj04 (approved):
Lovely!

PR #201, opened 2023-05-10
Add ability to build a Judgment object from a known set of values

jacksonj04:
At the moment a Judgment can only be initialised from a URI and it will get other values as necessary, but the search flow returns a more complete set of data for each result.

We should be able to initialise a Judgment object from these pieces of data, which means retooling how a Judgment stores properties.

This needs to happen before we can implement search in the API layer.

### DIFF
--- a/src/caselawclient/models/judgments.py
+++ b/src/caselawclient/models/judgments.py
@@ -110,6 +110,12 @@ class Judgment:
 
         return self.values_store[attr_name]
 
+    def __setattr__(self, attr_name: str, value: MarkLogicValueType) -> None:
+        if attr_name not in self.marklogic_attribute_map:
+            raise AttributeError
+
+        self.__dict__["values_store"][attr_name] = value
+
     @property
     def public_uri(self) -> str:
         return "https://caselaw.nationalarchives.gov.uk/{uri}".format(uri=self.uri)
--- a/src/caselawclient/models/judgments.py
+++ b/src/caselawclient/models/judgments.py
@@ -1,6 +1,6 @@
 import datetime
 from functools import cached_property
-from typing import Optional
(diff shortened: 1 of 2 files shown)


src/caselawclient/models/judgments.py
```diff
@@ -23,66 +23,101 @@
...
+    # These type definitions make sure that downstream users of this class can rely on the types
+    uri: str
+    api_client: MarklogicApiClient
+    name: str
+    neutral_citation: str
+    court: str
+    judgment_date_as_string: str
+    is_published: bool
+    is_sensitive: bool
+    is_anonymised: bool
```
anthonyhashemi, line 49: Isn't it only `MARKLOGIC_ATTRIBUTE_MAP`, `values_store`, `uri` and `api_client` that are properties on the `Judgment` class now, and that therest are just keys in the `MARKLOGIC_ATTRIBUTE_MAP`?

src/caselawclient/models/judgments.py
```diff
@@ -23,66 +23,101 @@
...
         if not self.judgment_exists():
             raise JudgmentNotFoundError(f"Judgment {self.uri} does not exist")
 
+        # Callables for setter accept a single argument of the judgment URI
+        self.MARKLOGIC_ATTRIBUTE_MAP = {
+            "name": {
+                "getter": self.api_client.get_judgment_name,
+                "setter": None,
+            },
+            "neutral_citation": {
```
anthonyhashemi, line 63: I don't think I'm quite grasping what's going on with this attr override.

Quick call after standup to go through it?

jacksonj04:
Closing in favour of work to store/manipulate XML directly within a `Document`.

PR #205, opened 2023-05-12
Add optional collections kwarg to Client.advanced_search  method for …

anthonyhashemi:
## Changes in this PR:

- Added optional `collections` kwarg to `Client.advanced_search` method for filtering by collections
- improved the tests for `Client.advanced_search`
- Changelog updated appropriately

## Testing
- [x] Before merging, I need to test the client manually to show that it is working end to end especially since the python tests mock out the calls to marklogic.

UPDATE: Have tested locally by instantiating `MarklogicAPIClient` with credentials pointing to our staging marklogic instance, added some documents manually to some collections in the marklogic console and then called `client.advanced_search("judgment")` and verified that it returned those documents from that collection.

### DIFF
--- a/tests/client/test_advanced_search.py
+++ b/tests/client/test_advanced_search.py
@@ -3,6 +3,8 @@ import logging
 import unittest
 from unittest.mock import patch
 
+import pytest
+
 from caselawclient.Client import MarklogicApiClient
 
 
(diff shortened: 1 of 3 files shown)


dragon-dxw (approved):
Thanks for the shiny new tests!

tests/client/test_advanced_search.py
```diff
@@ -3,18 +3,126 @@
...
+        """
+        Scenario: MarkLogic error while searching
+        Given a client instance
+        When the advanced_search method is called and MarkLogic returns an error
+        Then it should raise an exception with the error message
+        """
+        error_message = "Error message from MarkLogic"
+        with patch.object(self.client, "invoke"):
+            self.client.invoke.side_effect = Exception(error_message)
+            with pytest.raises(Exception) as e:
```
dragon-dxw, line 113: Can we say anything useful about the return type of the test? (It's the same / it's not the same)

Consider using RuntimeError rather than the more generic Exception group?

anthonyhashemi: Any exception that invoke raises will be raised in this function as there is no try catch. I have updated the test so that it asserts that the same exception object is raised though!

PR #207, opened 2023-05-16
Make collections parameter a list in advanced_search

anthonyhashemi:
When filtering by judgment collection in the PUI in https://github.com/nationalarchives/ds-caselaw-public-ui/pull/701/files, I realised it was nicer for `collections` to be a list rather than a comma-separated string - so this PR changes that.

## Changes in this PR:
- Adjust collections filter on advanced_search so that it is passed as a list of possible collection uri strings instead of a comma separated list in a string
- Added some empty string/space filtering to it, so that it doesn't error in the marklogic query due to spaces/empty strings as marklogic collection names must be valid uris and empty strings / strings with spaces are not valid. Could do more validation but I dont think its worth it for now and probably best to do in the marklogic query itself. These filters seemed simple enough to add though, so I did. Thoughts?

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -501,7 +501,9 @@ class MarklogicApiClient:
                 "to": str(date_to or ""),
                 "show_unpublished": str(show_unpublished).lower(),
                 "only_unpublished": str(only_unpublished).lower(),
-                "collections": str(collections or ""),
+                "collections": ",".join(collections or [])
+                .replace(" ", "")
+                .replace(",,", ","),
             }
         )
         return self.invoke(module, vars)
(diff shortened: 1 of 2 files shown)


dragon-dxw:
Good spot -- have you considered using 
`declare variable $court as json:array? external;` in the xquery 
https://github.com/nationalarchives/ds-caselaw-marklogic/blob/0d7454395c7ce6a36ea8f0f713cea1ba4e1764f0/src/main/ml-modules/root/judgments/search/search-v2.xqy#L17
and passing an actual list from python?
https://github.com/nationalarchives/ds-caselaw-custom-api-client/blob/11d37bc34e556d7cefee914feda096917b56fdb2/src/caselawclient/Client.py#LL491C3-L491C3

dragon-dxw (approved):

PR #211, opened 2023-05-19
Abstract advanced_search_results and SearchResults from PUI and EUI into apiclient

anthonyhashemi:
## Changes in this PR:
- Added `SearchParameters` dataclass for use with search functions using the legacy kwargs from `Client.advanced_search` and new `collections` for filtering by collections from https://github.com/nationalarchives/ds-caselaw-custom-api-client/pull/205
- Changed `Client.advanced_search` interface to take in `SearchParameters` as opposed to the legacy kwargs.
- Added `search_and_parse_results` ( `advanced_search_results` in PUI and EUI) and `search_judgments_and_parse_results` methods to `Client`
- Added `SearchResults` class to encapsulate document search results

More details about these additions /changes in comments on code in PR.

Changelog updated.


tests/client/test_search_and_parse_results.py
```diff
@@ -0,0 +1,104 @@
+from unittest.mock import MagicMock, patch
+
+from lxml import etree
+
+from caselawclient.Client import SearchParameters, api_client
+from caselawclient.models.search_results import SearchResults
+
+
+@patch("caselawclient.Client.api_client.advanced_search")
+def test_perform_advanced_search(mock_advanced_search):
```
anthonyhashemi, line 10: Taken this from the current open PR in public ui which I will be updating once this PR is merged in and new api client is released:

https://github.com/nationalarchives/ds-caselaw-public-ui/pull/701/files#diff-95daf164b3b15f1fb2437325b5b44436775eaa1a6c976142426f1247455d3080

src/caselawclient/Client.py
```diff
@@ -38,6 +41,24 @@
 DEFAULT_XSL_TRANSFORM = "accessible-html.xsl"
 
 
+@dataclass
+class SearchParameters:
```
anthonyhashemi, line 45: Decided to create `SearchParameters` dataclass so that:

- we dont have to specify all kwargs all the way up the search chain
- in turn we dont have to make sure we keep the correct default values of kwargs up the chain - just define it in one place here

src/caselawclient/Client.py
```diff
@@ -103,21 +127,29 @@ def _path_to_request_url(self, path: str) -> str:
     def _court_list_splitter(self, court_text: str) -> Set[str]:
         return set(court_text.lower().replace(" ", "").split(","))
 
-    def _court_list(self, court_text: str) -> Optional[list[str]]:
-        if not court_text.strip():
-            return None
-        alt_names = {
+    def _join_court_text(self, court_text: str) -> str:
```
anthonyhashemi, line 130: only new thing for the court_list method is `_join_court_text` which has been taken from public_ui's usage of `api_client.advanced_search`

`https://github.com/nationalarchives/ds-caselaw-public-ui/blob/6c966bba962e3a8efdd951029b68fa9cb576b3e5/judgments/utils.py#L42`

src/caselawclient/Client.py
```diff
@@ -103,21 +127,29 @@ def _path_to_request_url(self, path: str) -> str:
...
+        return ",".join(court_text) if isinstance(court_text, list) else court_text
+
+    def _get_alternative_court_names(self, courts: Set[str]) -> Set[str]:
```
anthonyhashemi, line 133: just some refactoring

src/caselawclient/Client.py
```diff
@@ -448,23 +480,7 @@ def invoke(
...
-        order: Optional[str] = None,
-        date_from: Optional[str] = None,
-        date_to: Optional[str] = None,
-        page: int = 1,
-        page_size: int = RESULTS_PER_PAGE,
-        show_unpublished: bool = False,
-        only_unpublished: bool = False,
-        collections: Optional[str] = None,
-    ) -> requests.Response:
+    def advanced_search(self, search_parameters: SearchParameters) -> requests.Response:
```
anthonyhashemi, line 483: just take in a SearchParameters object now for all the search functions in the search chain as explained on the comment at the SearchParameters definition

src/caselawclient/Client.py
```diff
@@ -753,6 +771,19 @@ def get_properties_for_search_results(self, judgment_uris: list[str]) -> str:
         response = self._send_to_eval(vars, "get_properties_for_search_results.xqy")
         return decode_multipart(response)
 
+    def search_and_parse_results(
```
anthonyhashemi, line 774: What we currently call `perform_advanced_search` in public and editor uis:

https://github.com/nationalarchives/ds-caselaw-public-ui/blob/6c966bba962e3a8efdd951029b68fa9cb576b3e5/judgments/utils.py#L42

and 

https://github.com/nationalarchives/ds-caselaw-editor-ui/blob/fd8f1e9100a8a60f081eecb79836f44688e70c94/judgments/utils/perform_advanced_search.py

src/caselawclient/Client.py
```diff
@@ -753,6 +771,19 @@ def get_properties_for_search_results(self, judgment_uris: list[str]) -> str:
...
+        self, search_parameters: SearchParameters
+    ) -> SearchResults:
+        response = self.advanced_search(search_parameters)
+        response_string = decode_multipart(response)
+        return SearchResults(response_string)
+
+    def search_judgments_and_parse_results(
```
anthonyhashemi, line 781: Had started updating `perform_advanced_search` in public ui to filter by `collections="judgment"`, but I think now we abstract this here we should keep a general `search_and_parse_results` (equivalent to perform_advanced_search`) but also create a judgment specific one as we currently only care about judgment - thus this method is created :)

src/caselawclient/Client.py
```diff
@@ -753,6 +771,19 @@ def get_properties_for_search_results(self, judgment_uris: list[str]) -> str:
...
+        self, search_parameters: SearchParameters
+    ) -> SearchResults:
+        search_parameters.collections = [JUDGMENT_COLLECTION_URI]
```
anthonyhashemi, line 784: `SearchParameters.collections` defaults to `None`
but here we override it to be `[JUDGMENT_COLLECTION_URI]` so that when this function is called we only return docs in the `JUDGMENT_COLLECTION_URI` collection (even if someone had specified the collections property in the `SearchParameters` input object.

src/caselawclient/models/search_results.py
```diff
@@ -0,0 +1,52 @@
+from typing import List
+
+from lxml import etree
+
+
+class SearchResults:
```
anthonyhashemi, line 6: Same

public ui: https://github.com/nationalarchives/ds-caselaw-public-ui/blob/6c966bba962e3a8efdd951029b68fa9cb576b3e5/judgments/models.py#L128

editor ui: https://github.com/nationalarchives/ds-caselaw-editor-ui/blob/fd8f1e9100a8a60f081eecb79836f44688e70c94/judgments/models/__init__.py#L143

but converter from using django-xml to just use lxml

Still need to do a followup PR where I move in SearchResult and SearchMatch - bit more complicated though and not entirely necessary for the search_results test so left for separate PR

src/caselawclient/Client.py
anthonyhashemi, line 774: @jacksonj04 @dragon-dxw 

At first, I wasn't sure about placing the 2 new methods `search_and_parse_results` `search_judgments_and_parse_results` in the Client library.

But after seeing that we have a mixture of methods, some that return:
- raw response objects
- text from decoded multipart responses
- other values given parsing of the decoded text

as well as the more boilerplatey eval, invoke etc methods, I though that although I think the class is probably doing too much at the moment, these 2 new methods do fit in with bulletpoint 3.

However, I do think at some point we should think about breaking this class out - I've had some thoughts on various options but think it needs some proper discussion.

src/caselawclient/Client.py
jacksonj04: I've never had to use `@dataclass` in anger but have been looking for an excuse - nice!

src/caselawclient/Client.py
jacksonj04: I've got some work in-flight at the moment which tackles this problem from the other end, ie "we should be able to take a bunch of stuff from a response and always get a coherent object or set of objects", so this will (I think) neatly meet up in the middle.

tests/client/test_search_and_parse_results.py
```diff
@@ -0,0 +1,104 @@
...
+
+    # Assert that SearchResults object was created successfully
+    assert isinstance(search_results, SearchResults)
+    assert search_results.total == "2"
+    assert len(search_results.results) == 2
+    assert search_results.results[0].attrib == {
+        "index": "1",
+        "uri": "/a/b/2011/10.xml",
+        "path": 'fn:doc("/a/b/2011/10.xml")',
+        "score": "40610",
```
dragon-dxw, line 85: I'm not sure I like this test. I don't know that we depend on the scope, confidence or fitness, the search result XML is pretty huge and we're broadly testing for "xml in = xml out". 

I'd drop the uncared for attribute tests, and heavily truncate the search results, or consider migrating this out to a test fixture.

dragon-dxw (approved):

tests/client/test_search_and_parse_results.py
anthonyhashemi: Yeah you make a good point, that is indeed what Im trying to show here - that xml in = xml out and those particular fields arent special in any way.

I had copied the response (that I am mocking) from an actual call to ML and modified it a bit so that it didnt refer to any real data, but you're right - I can definitely truncate the data a lot - so I'll give that a go!

src/caselawclient/Client.py
```diff
@@ -485,23 +501,25 @@ def advanced_search(
...
-                "page": max(1, int(page)),
-                "page-size": int(page_size),
-                "q": str(q or ""),
-                "party": str(party or ""),
-                "neutral_citation": str(neutral_citation or ""),
-                "specific_keyword": str(specific_keyword or ""),
-                "order": str(order or ""),
-                "from": str(date_from or ""),
-                "to": str(date_to or ""),
+                "court": self._court_list(search_parameters.court or ""),
```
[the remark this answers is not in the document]
anthonyhashemi: have created a `to_marklogic_data` method on the SearchParameters dataclass

PR #216, opened 2023-05-23
Add SearchResult dataclass to api client library

anthonyhashemi:
## Changes in this PR:
- Added `SearchResult` and `SearchMatches` and `SearchResultMeta` due to being shared between PUI AND EUI

Differences to pui and eui implmentations:
- `SearchResultMeta` was added - but was found only in editor ui. 
- As part of this, note that publicui currently has `author` in main `SearchResult` class and note that it also has `last_modified` whereas not included in `SearchResult` or `SearchResultMeta` in editor ui so added to `SearchResultMeta` here.
- Made `SearchResult` and `SearchResultMeta` dataclasses. moving any dataprocessing logic from init methods into the static factory methods.
- `SearchMatches` not using django-xml as the client isnt a django project - uses lxml directly.
- Added tests for all classes!

- Have updated the changelog!

### PUI
- `SearchResult` and `SearchMatch`: https://github.com/nationalarchives/ds-caselaw-public-ui/blob/8b9b5aa1101adfae58d50f7cb3c74e3b4c9be9c0/judgments/models.py#L17

### EUI
- `SearchResult`, `SearchMatch` and `SearchResultMeta`: https://github.com/nationalarchives/ds-caselaw-editor-ui/blob/a736c373a793c70794563b861781dc8fcb029d31/judgments/models/__init__.py#L73


src/caselawclient/search_result.py
```diff
@@ -0,0 +1,264 @@
...
+    Return the first item in the given iterable.
+
+    Args:
+        x (Any): The iterable.
+
+    Returns:
+        Optional[Any]: The first item in the iterable or None if the iterable is empty.
+    """
+    if len(x) > 1:
+        raise RuntimeError("More than one found.")
```
anthonyhashemi, line 261: this function was taken straight from PUI and EUI as well.
as was  the comment `# we assume that there is only one node in the search result XML.` on line  233

This runtimerror line is not currently covered by any tests... 

I will probably rename and rewite this function so that it is clearer that it is concerned with when an xpath query returns more than 1 result.

Then will write a test for it

src/caselawclient/search_result.py
```diff
@@ -0,0 +1,264 @@
...
+                neutral_citation=neutral_citation,
+                name=metadata_name,
+                matches=matches_html,
+                court=court,
+                date=date,
+                content_hash=content_hash,
+                transformation_date=transformation_date,
+                meta=meta,
+            )
+        except caselawclient.errors.MarklogicAPIError as e:
```
anthonyhashemi, line 134: EditorUI had the try except whereas PUI didnt.

I imagine we saw a bug in EUI and thats why we added that.


Will add a test unless people think it is unncessary to have this try except

src/caselawclient/search_result.py
```diff
@@ -0,0 +1,264 @@
...
+            author_email=one(node.xpath("//source-email/text()")) or "",
+            consignment_reference=one(
+                node.xpath("//transfer-consignment-reference/text()")
+            )
+            or "",
+            submission_datetime=submission_datetime,
+            assigned_to=one(node.xpath("//assigned-to/text()")) or "",
+            editor_hold=one(node.xpath("//editor-hold/text()")) or "",
+            editor_priority=one(node.xpath("//editor-priority/text()"))
+            or "20",  # medium priority
```
anthonyhashemi, line 245: feel like this probably should come from an Enum somewhere - where do we define these priority numbers?

src/caselawclient/search_result.py
```diff
@@ -0,0 +1,264 @@
...
+    author_email: str = ""
+    consignment_reference: str = ""
+    submission_datetime: Optional[datetime] = None
+    assigned_to: str = ""
+    editor_hold: str = "false"
+    editor_priority: str = ""
+    last_modified: str = ""
+
+    @property
+    def editor_status(
```
anthonyhashemi, line 180: In EUI need to do `item.meta.editor_status.value` instead of `item.meta.editor_status.value`

src/caselawclient/search_result.py
```diff
@@ -0,0 +1,264 @@
...
+            )
+            return None
+
+
+class EditorStatus(Enum):
+    """
+    Enum representing the editor status.
+    """
+
+    HOLD = "hold"
```
anthonyhashemi, line 146: probably worth reordering to
NEW, IN_PROGRESS, HOLD

CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog 1.0.0].
 - Added `SearchParameters` dataclass for use with search functions using the legacy kwargs from `Client.advanced_search` and new `collections` field for filtering by collections
 - Changed `Client.advanced_search` interface to take in `SearchParameters` as opposed to the legacy kwargs.
 - Added `search_and_parse_results` and `search_judgments_and_parse_results` methods to `Client`
-- Added `SearchResults` class to encapsulate document search results
+- Added `SearchResults`, `SearchResult`, `SearchResultMeta`, `SearchMatches` class to encapsulate and process document search result
```
anthonyhashemi, line 11:
```suggestion
- Added `SearchResults`, `SearchResult`, `SearchResultMeta`, `SearchMatches` class to encapsulate and process document search results
```

src/caselawclient/search_result.py
```diff
@@ -0,0 +1,264 @@
...
+    Attributes:
+        uri (str): The URI of the search result.
+        neutral_citation (str): The neutral citation of the search result.
+        name (str): The name of the search result.
+        court (str): The court of the search result.
+        date (datetime): The date of the search result.
+        matches (str): The search result matches.
+        content_hash (str): The content hash of the search result.
+        transformation_date (str): The transformation date of the search result.
+        meta (SearchResultMeta): The metadata of the search result.
```
anthonyhashemi, line 31:
```suggestion
        meta (Optional[SearchResultMeta]): The metadata of the search result.
```

src/caselawclient/search_result.py
anthonyhashemi: removed this function and refactored all the xpath code to simplify

src/caselawclient/search_result.py
anthonyhashemi: Remove the try except -it looked like:
```
        try:
            return SearchResult(
                uri=uri,
                neutral_citation=neutral_citation,
                name=metadata_name,
                matches=matches_html,
                court=court,
                date=date,
                content_hash=content_hash,
                transformation_date=transformation_date,
                meta=meta,
            )
        except caselawclient.errors.MarklogicAPIError as e:
            logging.warning(
                f"Unable to create search result for {uri}. Has it been deleted? Full error: {e}"
            )
            return None
```

but since I have refactored the metatadata initialisation out of this class, it is no longer needed here. Considered puttin in SearchResultsMeata.create_from_node around the apiclient calls but then I thought that the try except was definitely particular to the implementation in EUI and that the client library shouldnt decide to return a None object when SearchResultsMeta is called. If this does error, we can continue handling in eui (and maybe in PUI too).

src/caselawclient/search_result.py
```diff
@@ -0,0 +1,226 @@
...
+        """
+        Create a SearchResultMeta instance from a URI.
+
+        Args:
+            uri (str): The URI of the search result.
+
+        Returns:
+            SearchResultMeta: The created SearchResultMeta instance.
+        """
+        response_text = api_client.get_properties_for_search_results([uri])
```
anthonyhashemi, line 185: The SearchResultMeta class is currently passed to the SearchResult class through dependency injection.

This is because it needs to make calls to marklogic with `api_client`.

In my mind this is just a stop gap, but ideally I feel like we could return the "metadata" of a document as part of the `api_client.advanced_search`  (Marklogic `search_v2.xqy`) call. At the moment, to parse the results, we are having to make n extra calls for every advanced_search call where n is the number of results in the search which is inefficient. On top of that, we wouldnt need this extra code!

What do you guys think?

@dragon-dxw @jacksonj04

anthonyhashemi:
Closing this PR in favour of opening 2 smaller prs due to a change in the abstraction that was sparked after chat with Nick and Dragon

anthonyhashemi:
new PRs:
- https://github.com/nationalarchives/ds-caselaw-custom-api-client/pull/222
- https://github.com/nationalarchives/ds-caselaw-custom-api-client/pull/223

PR #236, opened 2023-06-06
Distinguish documents at ingestion

anthonyhashemi:
## Changes in this PR:
Distinguish documents at ingestion

## Manual Testing of the xquery:
Screenshot 2023-06-06 at 18 10 12


src/caselawclient/Client.py
```diff
@@ -327,18 +327,18 @@ def save_judgment_xml(
 
         return self._send_to_eval(vars, "update_judgment.xqy")
 
-    def insert_judgment_xml(
-        self, judgment_uri: str, judgment_xml: Element
+    def insert_document_xml(
```
anthonyhashemi, line 330: generalising the name to document since we now have press summaries too

src/caselawclient/xquery/insert_document.xqy
```diff
@@ -0,0 +1,15 @@
...
+
+import module namespace dls = "http://marklogic.com/xdmp/dls" at "/MarkLogic/dls.xqy";
+
+declare variable $uri as xs:string external;
+declare variable $document as xs:string external;
+
+let $document_xml := xdmp:unquote($document)
+
+let $collections :=
+  if (fn:contains($uri, 'press_summary'))
```
anthonyhashemi, line 11: we decided that we will have `/press_summary/` in the uri, so this is a good pattern to look for to distinguish the doc types on ingestion

tests/client/test_save_copy_delete_judgment.py
```diff
@@ -71,19 +71,19 @@ def test_save_locked_judgment_xml_checks_content_hash(self):
...
                 self.client.save_locked_judgment_xml(uri, judgment_xml, "my annotation")
 
-    def test_insert_judgment_xml(self):
+    def test_insert_document_xml(self):
         with patch.object(self.client, "eval") as mock_eval:
-            uri = "/ewca/civ/2004/632"
-            judgment_str = "<root>My new judgment</root>"
-            judgment_xml = ElementTree.fromstring(judgment_str)
+            uri = "/ewca/civ/2004/632/"
+            document_str = "<root>My judgment</root>"
```
anthonyhashemi, line 77: just generalising again

src/caselawclient/xquery/insert_document.xqy
jacksonj04: This makes sense for now; I'd like to start relying on a structured data bundle arriving from the parser in future so we can move all the things like failure states into there as well, but that's a future step.

tests/client/test_save_copy_delete_judgment.py
```diff
@@ -71,19 +71,19 @@ def test_save_locked_judgment_xml_checks_content_hash(self):
...
                 "uri": "/ewca/civ/2004/632.xml",
-                "judgment": judgment_str,
+                "document": document_str,
             }
-            self.client.insert_judgment_xml(uri, judgment_xml)
+            self.client.insert_document_xml(uri, document_xml)
 
             assert mock_eval.call_args.args[0] == (
-                os.path.join(ROOT_DIR, "xquery", "insert_judgment.xqy")
+                os.path.join(ROOT_DIR, "xquery", "insert_document.xqy")
```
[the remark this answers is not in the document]
jacksonj04: For now, run with this as-is. I'd like to move as much of the xquery as possible into the ML repo and get it loaded server-side rather than being sent on the wire each time, but that can happen separately and then we only need to worry about xquery tests in one place (and just verify that we're calling the right ML function in the API client)

jacksonj04 (approved):
:shipit:

PR #239, opened 2023-06-07
Improve timeout error

dragon-dxw:
* Don't raise an unrelated error on a timeout
* Ensure that the error handling has a test
* Update Poetry in general.
* Pre-commit was playing up -- feels like CI was already using 1.3.0...
* Adds dev dependency on "responses", a Requests mocking library.


tests/client/test_client.py
```diff
@@ -1,7 +1,34 @@
...
+        with responses.RequestsMock() as response_list:  # type: ignore[attr-defined]
+            response_list.add(
+                responses.GET,  # type: ignore[attr-defined]
+                url="http://example.com",
+                status=504,
+                body="Example Gateway Timeout.",
+            )
+            response = requests.get("http://example.com")
+
+        # response = MagicMock()
```
anthonyhashemi, line 26: can we remove this commented code

tests/client/test_client.py
```diff
@@ -1,7 +1,34 @@
...
+        # response.raise_for_status.side_effect=HTTPError(response=response)
+        # response.status_code=504
+        with pytest.raises(GatewayTimeoutError) as gateway_exception:
+            self.client._raise_for_status(response)
```
anthonyhashemi, line 30: Not for now, but this got me thinking, maybe we could create a `MarklogicResponse` class that looks like

````
class MarkLogicResponse(requests.Response):
    def raise_for_status(self):
        # .... current _raise_for_status implementation here ....
```

that way we would separate concerns a bit and thus no longer need to test a private method

anthonyhashemi (approved):
Looks good 👍

anthonyhashemi (approved):
Looks good 👍

PR #247, opened 2023-06-15
Introduce new Judgment.validation_failure_messages method

jacksonj04:
This method returns a list of reasons why the judgment has failed validation and is not publishable, by iterating over a list of attributes of a judgment which must 'pass' in order for it to be considered publishable and appending their error message if they fail. This also means we can refactor `is_publishable` to lean on the same list of attributes to test, which reduces function complexity.

Take this opportunity to tidy up the organisation of tests, and to also make parser failure a publish-blocking condition.

### DIFF
--- a/src/caselawclient/models/judgments.py
+++ b/src/caselawclient/models/judgments.py
@@ -171,18 +171,53 @@ class Judgment:
 
         return True
 
+    # attribute name, value which passes validation, error message
+    VALIDATION_ATTRIBUTES: list[tuple[str, bool, str]] = [
+        (
+            "is_failure",
+            False,
+            "This judgment has failed to parse",
+        ),
+        (
+            "is_held",
+            False,
+            "This judgment is currently on hold",
+        ),
+        (
+            "has_name",
+            True,
+            "This judgment has no name",
+        ),
+        (
+            "has_ncn",
+            True,
+            "This judgment has no neutral citation number",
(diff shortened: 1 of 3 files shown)


Terry-Price:
@jacksonj04 
Not sure if design of notification panel is just a placeholder or not. but we could set the current yellow panel up a bit more:
- use the red colour, feels like the message is code red. 
- replace the word ‘judgment’ with ‘document’ to keep it neutral/scalable. 
- See what the panel looks like if it only was the width of the main content form below it. e.g. not right across the page. i’m not by that fussed about this last one, just might feel less aggressive when it’s in red.

src/caselawclient/models/judgments.py
```diff
@@ -171,19 +171,56 @@ def has_court(self) -> bool:
...
     def is_publishable(self) -> bool:
-        if (
-            self.is_held
-            or not self.has_name
-            or not self.has_ncn
-            or not self.has_valid_ncn
-            or not self.has_court
-        ):
-            return False
+        for function_name, pass_value, _ in self.VALIDATION_ATTRIBUTES:
```
dragon-dxw, line 210: Is it worth this just being `return not self.validation_failure_messages()` with an explanatory comment, so the code is not duplicated?

dragon-dxw (approved):

PR #265, opened 2023-06-29
Check for a valid court, rather than an present one

dragon-dxw:
We check if the court string is valid before allowing publication by checking against the CourtRepository in utils.


dragon-dxw:
[Invalid courts should be a publish-blocking condition](https://trello.com/c/UaDxjPo2/1050-invalid-courts-should-be-a-publish-blocking-condition)

anthonyhashemi (approved):
🚀

PR #294, opened 2023-07-10
Overwrite move judgment

dragon-dxw:
[not ready, left as stub]
Begin moving overwrite and move judgment so that they're in the custom api client and can be used by both editor and ingester.


dragon-dxw:
[Migrate save/versioning behaviour to the APIclient instead ingester/EUI/enrich having their own version](https://trello.com/c/TjKjCqZD/1108-migrate-save-versioning-behaviour-to-the-apiclient-instead-ingester-eui-enrich-having-their-own-version)

tests/client/test_client.py
```diff
@@ -14,9 +14,9 @@ def setUp(self):
         self.client = MarklogicApiClient("", "", "", False)
 
     def test_timeout(self):
-        with responses.RequestsMock() as response_list:  # type: ignore[attr-defined]
```
jacksonj04, line 17: \o/

src/caselawclient/models/utilities/move.py
```diff
@@ -0,0 +1,142 @@
...
+class MoveJudgmentError(Exception):
+    pass
+
+
+def overwrite_document(
+    source_uri: str,
+    target_citation: str,
+    api_client: Any,
+) -> str:
+    """Move the judgment at old_uri on top of the new citation, which must already exist
```
anthonyhashemi, line 28:
```suggestion
    """Move the document at old_uri on top of the new citation, which must already exist
```

src/caselawclient/models/utilities/move.py
```diff
@@ -0,0 +1,142 @@
...
+        raise RuntimeError(f"trying to overwrite yourself, {source_uri}")
+    if new_uri is None:
+        raise NeutralCitationToUriError(
+            f"Unable to form new URI for {source_uri} from neutral citation: {target_citation}"
+        )
+    if not api_client.document_exists(new_uri):
+        raise OverwriteJudgmentError(
+            f"The URI {new_uri} generated from {target_citation} does not already exist, so cannot be overwritten"
+        )
+    old_judgment = Judgment(source_uri, api_client)
```
anthonyhashemi, line 49:
```suggestion
    old_document = api_client.get_document_by_uri_or_404(source_uri)
```

src/caselawclient/models/utilities/move.py
```diff
@@ -0,0 +1,142 @@
...
+    try:
+        old_judgment_bytes = old_judgment.content_as_xml
+        old_judgment_xml = ET.XML(bytes(old_judgment_bytes, encoding="utf-8"))
```
anthonyhashemi, line 52: can we just call `old_document.content_as_xml`?

src/caselawclient/models/utilities/move.py
```diff
@@ -0,0 +1,142 @@
...
+        api_client.save_judgment_xml(
+            new_uri,
+            old_judgment_xml,
+            annotation=f"overwritten from {source_uri}",
+        )
+        set_metadata(source_uri, new_uri, api_client)
+        # TODO: consider deleting existing public assets at that location
```
anthonyhashemi, line 59: If we unpublish the published document using the `document.unpublish` method, this should be handled already.

src/caselawclient/models/utilities/move.py
```diff
@@ -0,0 +1,142 @@
...
+        api_client.delete_judgment(source_uri)
+    except MarklogicAPIError as e:
+        raise OverwriteJudgmentError(
+            f"Failure when attempting to delete judgment from {source_uri}: {e}"
+        )
+
+    return new_uri
+
+
+def update_document_uri(source_uri: str, target_citation: str, api_client: Any) -> str:
```
anthonyhashemi, line 77: isn't `overwrite_document_uri` essentially a superset of what update_document_uri can do?

i.e `update_document_uri` is redundant?

src/caselawclient/models/utilities/move.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    Move the document at old_uri to the correct location based on the neutral citation
```
anthonyhashemi, line 79:
```suggestion
    Move the document at source_uri to the correct location based on the neutral citation
```

src/caselawclient/models/utilities/aws.py
```diff
@@ -146,3 +151,40 @@ def notify_changed(uri: str, status: str, enrich: bool = False) -> None:
         Subject=f"Updated: {uri} {status}",
         MessageAttributes=message_attributes,
     )
+
+
+def copy_assets(old_uri: str, new_uri: str) -> None:
```
anthonyhashemi, line 156:
```suggestion
def copy_unpublished_assets(old_uri: str, new_uri: str) -> None:
```

rename based off the docstring which says `Copy *unpublished* assets...`

src/caselawclient/models/utilities/aws.py
```diff
@@ -146,3 +151,40 @@ def notify_changed(uri: str, status: str, enrich: bool = False) -> None:
...
+    bucket = env("PRIVATE_ASSET_BUCKET")
+    old_uri = uri_for_s3(old_uri)
+    new_uri = uri_for_s3(new_uri)
+
+    response = client.list_objects(Bucket=bucket, Prefix=old_uri)
+
+    for result in response.get("Contents", []):
+        old_key = str(result["Key"])
+        new_key = build_new_key(old_key, new_uri)
+        if new_key is not None:
```
anthonyhashemi, line 171: we can reduce nesting here:

```suggestion
        if new_key is None:
            continue
```

and then unident the code block that is currently in the if statement

anthonyhashemi (commented):
Having looked through the source code, I think this makes sense. Before I go over the tests, let's have a chat like you suggested.

anthonyhashemi (approved):
Looks good. If we find there's any issues when integrating into the repos that use we can always tweak this

PR #300, opened 2023-07-13
Add new `get_document_by_uri` method to API client

jacksonj04:


src/caselawclient/Client.py
```diff
@@ -93,6 +95,9 @@ def __init__(
         self.session = requests.Session()
         self.session.auth = HTTPBasicAuth(username, password)
 
+    def get_document_by_uri(self, uri: str) -> Document:
```
anthonyhashemi, line 98: I imagine a PressSummary class is coming in the next PR where this method will be extended to be conditional based on the document's collection?

anthonyhashemi (approved):

src/caselawclient/Client.py
jacksonj04: That's the plan; this is a short hop we can use to validate this behaviour in EUI/PUI and break them out of initialising a `Judgment` directly, and then PressSummary can arrive independently.

PR #303, opened 2023-07-17
`get_document_by_uri` now returns unique types for `Judgment` and `PressSummary`

jacksonj04:
Adds a new `PressSummary` class and make `get_document_by_uri` return the correct document type class. This makes use of a new `get_document_type_from_uri` method which checks a document's collections in MarkLogic and sees if any of them match a known collection for document types


anthonyhashemi (approved):

src/caselawclient/models/judgments.py
```diff
@@ -3,8 +3,8 @@
 
 from ds_caselaw_utils import neutral_url
 
-from caselawclient.models.documents import Document
-from caselawclient.xml_helpers import get_xpath_match_string
+from ..xml_helpers import get_xpath_match_string
```
anthonyhashemi, line 6: why change this to a relative path?

anthonyhashemi (approved):

PR #309, opened 2023-07-21
💔 [BREAKING CHANGE] Remove supplemental/anonymous/sensitive getters/setters 

dragon-dxw:
* Marklogic config doesn't touch these things -- getter/setter was generic boolean fields
* Ingester doesn't know about these things and does not explicitly set/reset them.
* Editor has already been tweaked to not get/set these things.

This is a BREAKING CHANGE to the API, and should be a major version when released.

https://trello.com/c/o2sNyDZ2/1170-remove-deprecated-sensitive-supplemental-anonymised-method-calls-from-api-and-marklogic


jacksonj04 (approved):
:fire:

PR #312, opened 2023-07-24
💔 [BREAKING] Better MarkLogic multipart response handling

jacksonj04:
This entire client class currently assumes that there is exactly one item coming from MarkLogic in a multipart response, or failing that the only useful part of the response is the first item. This is currently true, but will not necessarily be in the future (and is prone to flakiness in any case).

This moves handling of the multipart response and turning it into text into a distinct function which can be used in future to retrieve a list of results, and simplifies the existing `decode_multipart` function to return only the first result, or fail with a hard exception (because it would mean whatever is calling the method is expecting a single result in a context where multiple results have been returned).

This is potentially a **breaking change** since unexpected multiple returns now raise an exception.

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -46,22 +46,33 @@ class MultipartResponseLongerThanExpected(Exception):
     pass
 
 
+class NoStringResponseWhenExpected(Exception):
+    pass
+
+
 def get_multipart_strings_from_marklogic_response(
     response: requests.Response,
 ) -> list[str]:
-    # TODO: This is horrible, and should either return a None, or throw an exception.
     if not (response.content):
-        return [""]
+        return []
 
     multipart_data = decoder.MultipartDecoder.from_response(response)
 
     return [part.text for part in multipart_data.parts]
 
 
-def get_single_string_from_marklogic_response(response: requests.Response) -> str:
+def get_single_string_from_marklogic_response(
+    response: requests.Response,
+) -> str:
     parts = get_multipart_strings_from_marklogic_response(response)
     part_count = len(parts)
-    if part_count > 1:
+
+    if part_count == 0:
+        # TODO: This should strictly speaking be None, but fixing this involves refactoring a lot of other stuff which
+        # relies on "" being falsy.
+        return ""
+
+    elif part_count > 1:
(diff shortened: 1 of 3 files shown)


src/caselawclient/Client.py
```diff
@@ -42,23 +42,31 @@
...
-    # Arguably, this should return None -- it occurs when there are no
-    # matching entries
+
+def get_multipart_strings_from_marklogic_response(
+    response: requests.Response,
+) -> list[str]:
+    # TODO: This is horrible, and should either return a None, or throw an exception.
     if not (response.content):
-        return ""
+        return [""]
```
dragon-dxw, line 54: Is it worth having a pure function which returns None and a wrapper which transforms it into `[""]`, so future code can be better?

src/caselawclient/Client.py
```diff
@@ -42,23 +42,31 @@
...
 
     multipart_data = decoder.MultipartDecoder.from_response(response)
-    part_count = len(multipart_data.parts)
+
+    return [part.text for part in multipart_data.parts]
+
+
+def get_single_string_from_marklogic_response(response: requests.Response) -> str:
+    parts = get_multipart_strings_from_marklogic_response(response)
```
dragon-dxw, line 62: I'm not sure these are "parts" any more. "strings" might be worse, though...

dragon-dxw (commented):
I think we need to address the crashing errors caused by multiple parts before they come up.

src/caselawclient/Client.py
jacksonj04: I'm definitely not happy with this as it stands, and would prefer returning a `None`. We could always make the `get_string` flavour return `""` in case of an empty response, just to reduce downstream breakage.

src/caselawclient/Client.py
```diff
@@ -42,23 +42,31 @@
...
+    part_count = len(parts)
     if part_count > 1:
-        logging.warning(
-            f"Throwing away multipart data ({part_count} items, expected 1)"
+        raise MultipartResponseLongerThanExpected(
```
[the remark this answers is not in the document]
jacksonj04: Yeah, I'm erring towards intentionally breaking it to force us to fix things. This is a good opportunity to use staging to kick tyres.

PR #313, opened 2023-07-24
We don't have an element named root before akomaNtoso

dragon-dxw:
There seems to be some confusion caused by the tests containing `<root>` elements that these appear in the actual judgments and the XQuery should include them -- we remove these elements, reduce dependency on the judgment element (but we still need to take a look at the `.xqy` files) and make some of the tests more realistic.

(We should probably think about having some sample XML files for testing.)

### DIFF
--- a/src/caselawclient/models/documents.py
+++ b/src/caselawclient/models/documents.py
@@ -83,14 +83,14 @@ class Document:
     @cached_property
     def name(self) -> str:
         return self._get_xpath_match_string(
-            "//akn:FRBRname/@value",
+            "/akn:akomaNtoso/akn:*/akn:meta/akn:identification/akn:FRBRWork/akn:FRBRname/@value",
             {"akn": "http://docs.oasis-open.org/legaldocml/ns/akn/3.0"},
         )
 
     @cached_property
     def court(self) -> str:
         return self._get_xpath_match_string(
-            "/root/akn:akomaNtoso/akn:judgment/akn:meta/akn:proprietary/uk:court/text()",
+            "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:court/text()",
             {
                 "uk": "https://caselaw.nationalarchives.gov.uk/akn",
                 "akn": "http://docs.oasis-open.org/legaldocml/ns/akn/3.0",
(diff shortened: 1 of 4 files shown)


src/caselawclient/models/documents.py
```diff
@@ -81,14 +81,14 @@ def public_uri(self) -> str:
     @cached_property
     def name(self) -> str:
         return self._get_xpath_match_string(
-            "//akn:FRBRname/@value",
+            "/akn:akomaNtoso/akn:*/akn:meta/akn:FRBRname/@value",
```
anthonyhashemi, line 84: ah my bad, not sure why I added root in the first place. Thanks for catching

src/caselawclient/models/documents.py
```diff
@@ -81,14 +81,14 @@ def public_uri(self) -> str:
...
             {"akn": "http://docs.oasis-open.org/legaldocml/ns/akn/3.0"},
         )
 
     @cached_property
     def court(self) -> str:
         return self._get_xpath_match_string(
-            "/root/akn:akomaNtoso/akn:judgment/akn:meta/akn:proprietary/uk:court/text()",
+            "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:court/text()",
```
anthonyhashemi, line 91: ah my bad, not sure why I added root in the first place. Thanks for catching

src/caselawclient/models/documents.py
anthonyhashemi, line 84: Is there a reason for adding the `/akn:akomaNtoso/...` prefix?

Does that fix a current judgment specific bug, or just preempting any issues we might have with both judgment and press summary?

src/caselawclient/models/documents.py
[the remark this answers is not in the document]
dragon-dxw: I think there might be other bits but we can take a look

src/caselawclient/models/documents.py
dragon-dxw: Specificity; there can be other FRBRnames in annexes, and we'd like to cut down the number of things we have multiple names for.

anthonyhashemi (approved):

PR #317, opened 2023-07-28
Expose ncn property on PressSummary class

anthonyhashemi:
## Changes in this PR:

Exposed ncn property on PressSummary class.

https://trello.com/c/Xt3AAMES/1175-25-api-client-expose-ncnofrelatedjudgment-property-on-presssummary-class

**Still need to update changelog**

### DIFF
--- a/src/caselawclient/models/press_summaries.py
+++ b/src/caselawclient/models/press_summaries.py
@@ -1,32 +1,22 @@
 from functools import cached_property
 from typing import Any
 
-from ds_caselaw_utils import neutral_url
-
+from caselawclient.models.neutral_citation_mixin import NeutralCitationMixin
 from caselawclient.xml_helpers import get_xpath_match_string
 
 from .documents import Document
 
 
-class PressSummary(Document):
+class PressSummary(NeutralCitationMixin, Document):
+    """
+    Represents a press summary document.
+    """
+
     document_noun = "press summary"
     document_noun_plural = "press summaries"
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
-        self.attributes_to_validate = self.attributes_to_validate + [
-            (
-                "has_ncn",
-                True,
-                "This {document_noun} has no neutral citation number",
-            ),
-            (
-                "has_valid_ncn",
-                True,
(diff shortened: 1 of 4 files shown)


src/caselawclient/models/press_summaries.py
```diff
@@ -1,6 +1,54 @@
...
+            ),
+        ]
+
+        super(PressSummary, self).__init__(*args, **kwargs)
+
+    @cached_property
+    def neutral_citation(self) -> str:
+        return get_xpath_match_string(
+            self.content_as_xml_tree,
+            "/akn:akomaNtoso/akn:doc/akn:mainBody/akn:p/akn:neutralCitation/text()",
```
anthonyhashemi, line 35: everything apart from the xpath line, i.e where we find the neutral citation in the xml is the same between this and the Judgment class.

I am contemplating moving most of the code to the Document class and just setting the xpath on each subclass. Then if we introduce new document types which we want to extend Document class from we can change things if needed

tests/models/test_press_summaries.py
```diff
@@ -0,0 +1,99 @@
...
+            ("[2022] UKUT 1 (IAC)", True),
+            ("[2022] EAT 1", True),
+            ("[2022] UKFTT 1 (TC)", True),
+            ("[2022] UKFTT 1 (GRC)", True),
+            ("[2022] EWHC 1 (KB)", True),
+            ("", False),
+            ("1604] EWCA Crim 555", False),
+            ("[2022 EWHC 1 Comm", False),
+            ("[1999] EWCOP", False),
+            ("[2022] UKUT B1 IAC", False),
```
anthonyhashemi, line 68: alot of duplicate tests

src/caselawclient/models/press_summaries.py
anthonyhashemi: have refactored so now have a NeutralCitationMixin class as per @jacksonj04 's suggestion on slack.

jacksonj04 (approved):
The docstrings could do with tweaking to match Sphinx syntax so pdoc can make them pretty, but otherwise :shipit:

PR #330, opened 2023-08-01
Bugfix: Ensure marklogic responses are treated as unencoded bytes

timcowlishaw:
## Changes in this PR:

Tiny bugfix release for a problem I uncovered when attempting to upgrade the editor UI to v13.0.0.

Marklogic returns strings with an encoding declaration, which are unsupported by lxml (which gives precedence to the encoding declaration in the xml itself - see [this stack overflow answer](https://stackoverflow.com/questions/28534460/lxml-etree-xml-valueerror-for-unicode-string)), and raises an ValueError with message "Unicode strings with encoding declaration are not supported. Please use bytes input or XML fragments without declaration." - this can be verified by attempting to access any judgment on staging markdown with version 13.0.0 of the client.

Instead, we can just preserve the marklogic response as bytes and pass them directly to lxml rather than decoding (and re-encoding later).

### DIFF
--- a/tests/models/test_documents.py
+++ b/tests/models/test_documents.py
@@ -182,6 +182,16 @@ class TestDocument:
             "test/1234", show_unpublished=True
         )
 
+    def test_judgment_content_as_xml_tree(self, mock_api_client):
+        mock_api_client.get_judgment_xml.return_value = (
+            b'<?xml version="1.0" encoding="UTF-8"?><xml></xml>'
+        )
+        try:
+            document = Document("test/1234", mock_api_client)
+            document.content_as_xml_tree
+        except ValueError as ex:
+            assert False, f"document.content_as_xml_tree raised a ValueError: {ex}"
+
     def test_document_status(self, mock_api_client):
         in_progress_document = Document("test/1234", mock_api_client)
         in_progress_document.is_held = False
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -70,6 +70,17 @@ def get_multipart_strings_from_marklogic_response(
     return [part.text for part in multipart_data.parts]
 
 
+def get_multipart_bytes_from_marklogic_response(
+    response: requests.Response,
+) -> list[bytes]:
+    if not (response.content):
+        return []
+
+    multipart_data = decoder.MultipartDecoder.from_response(response)
+
+    return [part.content for part in multipart_data.parts]
+
+
 def get_single_string_from_marklogic_response(
     response: requests.Response,
 ) -> str:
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog 1.0.0].
 
+## [Release 13.0.1]
```
jacksonj04, line 7: When updating the CHANGELOG you also need to update the links down at the very bottom of the document, it's not (yet) something which happens by magic.

tests/models/test_documents.py
```diff
@@ -182,6 +182,16 @@ def test_judgment_content_as_xml(self, mock_api_client):
             "test/1234", show_unpublished=True
         )
 
+    def test_judgment_content_as_xml_tree(self, mock_api_client):
+        mock_api_client.get_judgment_xml.return_value = (
+            '<?xml version="1.0" encoding="UTF-8"?><xml></xml>'
+        )
+        try:
```
jacksonj04, line 189: This is better written using [`with pytest.raises()`](https://docs.pytest.org/en/7.1.x/how-to/assert.html), which gives you a cleaner syntax.

jacksonj04 (changes requested):

anthonyhashemi:
I noticed that this branch is named v13.0.1. 

Just wanted to point out that this could confuse things with the current release process for this repo as outlined at the bottom of the readme.

Generally bug fixes /changes to code would be a normal PR, then we would make a release PR to update the version in the toml and changelog. I think the main reason for that is that that sort of PR can be approved really quickly and merged fast, unlike a pr with a real change which might only get merged in by the time another pr is already merged in.

anthonyhashemi:
> I noticed that this branch is named v13.0.1.
> 
> Just wanted to point out that this could confuse things with the current release process for this repo as outlined at the bottom of the readme.
> 
> Generally bug fixes /changes to code would be a normal PR, then we would make a release PR to update the version in the toml and changelog. I think the main reason for that is that that sort of PR can be approved really quickly and merged fast, unlike a pr with a real change which might only get merged in by the time another pr is already merged in.

@timcowlishaw  I'm going to make a release 13.1.0 this afternoon once my press summary ncn PR is in anyway, so no point releasing 13.0.1 alone. We can release both these changes together

timcowlishaw:
aha no probs @anthonyhashemi  i'll revert the version bump commit and rename the branch :-)

anthonyhashemi:
> aha no probs @anthonyhashemi i'll revert the version bump commit and rename the branch :-)

Ah I dont think you need to revert it. I just wanted to point things out!, I can sort out the changelog and versaion in the release_13.1.0 PR later!

tests/models/test_documents.py
timcowlishaw: aha, I had a look at this, but in the case where we're testing that an invocation does **not** raise an exception, I couldn't see a way to use `pytest.raises`? There's an alternative with conditional raising [described here](https://stackoverflow.com/questions/20274987/how-to-use-pytest-to-check-that-error-is-not-raised) but to my eyes that adds complexity rather than removing it

timcowlishaw:
Closing in favour of  #332 so as not to confuse the bugfix with the version bump

PR #332, opened 2023-08-01
Bugfix: Ensure marklogic responses are treated as unencoded bytes

timcowlishaw:
## Changes in this PR:

Tiny bugfix release for a problem I uncovered when attempting to upgrade the editor UI to v13.0.0.

Marklogic returns strings with an encoding declaration, which are unsupported by lxml (which gives precedence to the encoding declaration in the xml itself - see [this stack overflow answer](https://stackoverflow.com/questions/28534460/lxml-etree-xml-valueerror-for-unicode-string)), and raises an ValueError with message "Unicode strings with encoding declaration are not supported. Please use bytes input or XML fragments without declaration." - this can be verified by attempting to access any judgment on staging markdown with version 13.0.0 of the client.

Instead, we can just preserve the marklogic response as bytes and pass them directly to lxml rather than decoding (and re-encoding later).

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -81,6 +81,17 @@ def get_multipart_strings_from_marklogic_response(
     return [part.text for part in multipart_data.parts]
 
 
+def get_multipart_bytes_from_marklogic_response(
+    response: requests.Response,
+) -> list[bytes]:
+    if not (response.content):
+        return []
+
+    multipart_data = decoder.MultipartDecoder.from_response(response)
+
+    return [part.content for part in multipart_data.parts]
+
+
 def get_single_string_from_marklogic_response(
     response: requests.Response,
 ) -> str:
@@ -101,6 +112,25 @@ def get_single_string_from_marklogic_response(
         # TODO: This should strictly speaking be None, but fixing this involves refactoring a lot of other stuff which
         # relies on "" being falsy.
         return ""
+
+    elif part_count > 1:
+        raise MultipartResponseLongerThanExpected(
+            f"Response returned {part_count} multipart items, expected 1"
+        )
+
+    return parts[0]
+
+
+def get_single_bytestring_from_marklogic_response(
+    response: requests.Response,
+) -> bytes:
+    parts = get_multipart_bytes_from_marklogic_response(response)
+    part_count = len(parts)
+
+    if part_count == 0:
+        # TODO: This should strictly speaking be None, but fixing this involves refactoring a lot of other stuff which
+        # relies on "" being falsy.
+        return b""
 
     elif part_count > 1:
         raise MultipartResponseLongerThanExpected(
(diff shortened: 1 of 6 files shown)


timcowlishaw:
This is a new version of #330 , closed so that we can merge without the version bump and release

src/caselawclient/models/documents.py
```diff
@@ -191,7 +191,7 @@ def versions(self) -> list[VersionsDict]:
             return []
 
     @cached_property
-    def content_as_xml(self) -> str:
+    def content_as_xml(self) -> bytes:
```
dragon-dxw, line 194: This might impact public-ui and editor-ui since it's a used public method.

dragon-dxw (approved):

src/caselawclient/models/documents.py
timcowlishaw: Hey @dragon-dxw - I've left content_as_xml as is and added a content_as_xml_bytestring method - how does that look?

tests/models/test_documents.py
```diff
@@ -117,6 +117,16 @@ def test_judgment_content_as_xml(self, mock_api_client):
...
         )
 
+    def test_judgment_content_as_xml_tree(self, mock_api_client):
+        mock_api_client.get_judgment_xml_bytestring.return_value = (
+            b'<?xml version="1.0" encoding="UTF-8"?><xml></xml>'
+        )
+        try:
+            document = Document("test/1234", mock_api_client)
+            document.content_as_xml_tree
+        except ValueError as ex:
```
anthonyhashemi, line 127: What is this test is trying to demonstrate here, that calling document.content_as_xml_tree raises a ValueError?

If so, you can use pytest.raises instead of try, except.

```
mock_api_client.get_judgment_xml_bytestring.return_value = (
           b'<?xml version="1.0" encoding="UTF-8"?><xml></xml>'
)
document = Document("test/1234", mock_api_client)
with pytest.raises(ValueError):
           document.content_as_xml_tree
```

timcowlishaw: aha we *don't* expect it, that's the point - we're asserting that that error is *not* raised!

src/caselawclient/Client.py
```diff
@@ -110,6 +121,25 @@ def get_single_string_from_marklogic_response(
     return parts[0]
 
 
+def get_single_bytestring_from_marklogic_response(
```
anthonyhashemi, line 124: do we need all of this? could we not just encode the xml string into bytes if lxml is complaining about bytes?

timcowlishaw: aha this came from the chat on slack earlier - @jacksonj04 and @dragon-dxw felt it was best if lxml expects bytes to just pass the raw response through directly

anthonyhashemi: Ah right, i just wonder if doing a .encode() is sufficient - stops us from needing to add and test new functions that arent used otherwise!

tests/models/test_documents.py
timcowlishaw: aha ace, will amend!

PR #337, opened 2023-08-02
💔 Generalise set document metadata methods to deal with press summaries

anthonyhashemi:
## Changes in this PR:
- Generalised the set judgment metadata Client methods to set document metadata methods
- specifically for name, court and date

Haven't done it for neutral citation as it's a bit more complicated. I think I will leave that for a followup PR.

Have tested manually by copying and pasting into my local marklogic server. Have considered moving these xqy files into our marklogic repo so we can test them there, which I still think is worth doing but we have considered setting up some sort of suite of E2E tests to make sure things are actually working when hooked up together.

### 💔 Breaking change
The `set_judgment_name` and `set_judgment_court` methods have been renamed.

https://trello.com/c/lppwqc94/1169-api-client-tidy-up-allow-press-summary-metadata-name-court-date-and-related-ncn-to-be-edited

### DIFF
--- a/src/caselawclient/xquery/set_metadata_court.xqy
+++ b/src/caselawclient/xquery/set_metadata_court.xqy
@@ -2,14 +2,18 @@ xquery version "1.0-ml";
 
 declare namespace akn = "http://docs.oasis-open.org/legaldocml/ns/akn/3.0";
 declare namespace uk = "https://caselaw.nationalarchives.gov.uk/akn";
+
+let $proprietary-node := document($uri)/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:court
+let $court-node := $proprietary-node/uk:court
+
 declare function local:delete($uri)
 {
-   xdmp:node-delete(document($uri)/akn:akomaNtoso/akn:judgment/akn:meta/akn:proprietary/uk:court)
+   xdmp:node-delete($court-node)
 };
 declare function local:edit($uri, $content)
 {
    xdmp:node-replace(
-     document($uri)/akn:akomaNtoso/akn:judgment/akn:meta/akn:proprietary/uk:court,
+     ,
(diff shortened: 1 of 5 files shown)


src/caselawclient/xquery/set_metadata_name.xqy
```diff
@@ -4,15 +4,27 @@ declare namespace akn = "http://docs.oasis-open.org/legaldocml/ns/akn/3.0";
 declare variable $uri as xs:string external;
 declare variable $content as xs:string external;
 
-if (fn:boolean(
+declare function local:get-name-node($root-node)
```
dragon-dxw, line 7: I don't think this is necessary. We should be able to say `$document($uri)//akn:akomaNtoso/akn:*/akn:meta/...`, I think

anthonyhashemi: you were right, have simplified to this

src/caselawclient/xquery/set_metadata_court.xqy
```diff
@@ -2,22 +2,26 @@ xquery version "1.0-ml";
 
 declare namespace akn = "http://docs.oasis-open.org/legaldocml/ns/akn/3.0";
 declare namespace uk = "https://caselaw.nationalarchives.gov.uk/akn";
+
+let $proprietary-node := document($uri)/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:court
```
anthonyhashemi, line 6: importantly this now has `akn:*` rather than `akn:judgment` to allow for press summaries and eventually other docs types which have <doc> tag instead of <judgment> tag

dragon-dxw (approved):

dragon-dxw:
Consider whether leaving a deprecated set_judgment stub is appropriate, to avoid breaking changes that require immediate fixes.

PR #340, opened 2023-08-03
Best human identifier

dragon-dxw:
Add a "best_human_identifier" to all Document classes for display purposes

### DIFF
--- a/src/caselawclient/models/judgments.py
+++ b/src/caselawclient/models/judgments.py
@@ -28,3 +28,7 @@ class Judgment(NeutralCitationMixin, Document):
                 "akn": "http://docs.oasis-open.org/legaldocml/ns/akn/3.0",
             },
         )
+
+    @property
+    def best_human_identifier(self) -> str:
+        return self.neutral_citation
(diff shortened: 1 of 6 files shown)


src/caselawclient/models/press_summaries.py
```diff
@@ -27,3 +27,7 @@ def neutral_citation(self) -> str:
                 "akn": "http://docs.oasis-open.org/legaldocml/ns/akn/3.0",
             },
         )
+
+    @property
+    def best_human_identifier(self) -> Optional[str]:
```
jacksonj04, line 32: I don't think this is `Optional`, surely every summary has an NCN?

src/caselawclient/models/judgments.py
```diff
@@ -28,3 +28,7 @@ def neutral_citation(self) -> str:
                 "akn": "http://docs.oasis-open.org/legaldocml/ns/akn/3.0",
             },
         )
+
+    @property
+    def best_human_identifier(self) -> Optional[str]:
```
jacksonj04, line 33: Is this really `Optional`?

jacksonj04 (approved):
One query about if NCNs are `Optional`, otherwise 🚢

dragon-dxw:
I'd been fighting the typechecker, and you're absolutely right that I can just do that.

PR #345, opened 2023-08-08
Not found error for collections

dragon-dxw:
I'm not entirely happy with this.

Raising `fn:error(xs:QName("FCL_ERRORCODE"), "Description of Error"))` doesn't actually emit the error code in the 500 error that the API client receives. So we have to put that into the description.

The Python error currently is also not customised, looking like:

```E           caselawclient.errors.DocumentNotFoundError: 500 Server Error: Internal Server Error for url: ...:8011/LATEST/eval. Response body:
E           <error-response xmlns="http://marklogic.com/xdmp/error">
E             <status-code>500</status-code>
E             <status>Internal Server Error</status>
E             <message-code>FCL-DOCUMENTNOTFOUND No document at /not/a/real/url.xml</message-code>
E             <message>FCL-DOCUMENTNOTFOUND No document at /not/a/real/url.xml:</message>
E             <message-detail>
E               <message-title>FCL-DOCUMENTNOTFOUND No document at /not/a/real/url.xml</message-title>
E             </message-detail>
E           </error-response>
```

### DIFF
--- a/smoketest.py
+++ b/smoketest.py
@@ -18,7 +18,6 @@ api_client = Client.MarklogicApiClient(
 )
 
 
-@pytest.mark.xfail
 def test_get_document_type_returns_404():
     with pytest.raises(DocumentNotFoundError) as e:
         api_client.get_document_type_from_uri("/not/a/real/url")
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -29,6 +29,7 @@ from . import xquery_type_dicts as query_dicts
 from .content_hash import validate_content_hash
 from .errors import MarklogicAPIError  # noqa: F401
 from .errors import (
+    DocumentNotFoundError,
     GatewayTimeoutError,
     MarklogicBadRequestError,
     MarklogicCheckoutConflictError,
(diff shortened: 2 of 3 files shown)


smoketest.py
```diff
@@ -0,0 +1,22 @@
...
+    host=env("MARKLOGIC_HOST"),
+    username=env("MARKLOGIC_USER"),
+    password=env("MARKLOGIC_PASSWORD"),
+    use_https=env("MARKLOGIC_USE_HTTPS", default=None),
+)
+
+
+def test_get_document_type_returns_404():
+    with pytest.raises(DocumentNotFoundError) as e:
+        api_client.get_document_type_from_uri("/not/a/real/url")
```
anthonyhashemi, line 21: how does this test actually evaluate the the xquery when it is meant to be run in marklogic? Have we set up a marklogic instance to point it to? Are the env vars pointing to staging?

dragon-dxw: This is literally pointing at staging. It's not great, but it's at least something.

anthonyhashemi (approved):

PR #346, opened 2023-08-09
Remove `is_editable` in favour of `failed_to_parse`

jacksonj04:
The old `is_editable` property was very unclear. The new `failed_to_parse` property describes what it actually checks, as well as being tested.

As part of this, a document which has failed to parse is now unpublishable.


anthonyhashemi (approved):
Nice, much clearer 😎

PR #355, opened 2023-08-15
Fixed PressSummary `neutral_citation` property

anthonyhashemi:
## Changes in this PR:

- Fixed `neutral_citation` property to look within `preface` tag rather than `mainBody` for press summaries, due to updated parsing resulting in updated press summary xml structure.

https://trello.com/c/uJiohuAM/1269-eui-spelling-of-press-summary

### DIFF
--- a/src/caselawclient/models/press_summaries.py
+++ b/src/caselawclient/models/press_summaries.py
@@ -22,7 +22,7 @@ class PressSummary(NeutralCitationMixin, Document):
     def neutral_citation(self) -> str:
         return get_xpath_match_string(
             self.content_as_xml_tree,
-            "/akn:akomaNtoso/akn:doc/akn:mainBody/akn:p/akn:neutralCitation/text()",
+            "/akn:akomaNtoso/akn:doc/akn:preface/akn:p/akn:neutralCitation/text()",
             {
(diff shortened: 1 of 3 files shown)


src/caselawclient/models/press_summaries.py
```diff
@@ -22,7 +22,7 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
     def neutral_citation(self) -> str:
         return get_xpath_match_string(
             self.content_as_xml_tree,
-            "/akn:akomaNtoso/akn:doc/akn:mainBody/akn:p/akn:neutralCitation/text()",
+            "/akn:akomaNtoso/akn:doc/akn:preface/akn:p/akn:neutralCitation/text()",
```
dragon-dxw, line 25: Hmmm. Are there issues with moving press summaries because of this difference?

dragon-dxw (approved):
Looks good but highlights a can of worms.

src/caselawclient/models/press_summaries.py
anthonyhashemi: discussed offline - nope. we just need to update the `set_document_citation` method to handle updating ncns but we dont need that just yet. There is a card for that in TODO and I do have a branch where I started on it

PR #363, opened 2023-08-30
Add default user agent string to custom api client

dragon-dxw:
Set the default user agent string to give the version of the custom api client, and have it overriddable by users of the client.

### DIFF
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -1,3 +1,4 @@
+import re
 import unittest
 from types import SimpleNamespace
 from unittest.mock import MagicMock, patch
@@ -5,6 +6,7 @@ from unittest.mock import MagicMock, patch
 import pytest
 import requests
 import responses
+from requests import Request
 
 from caselawclient.Client import (
     MarklogicApiClient,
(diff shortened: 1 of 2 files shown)


dragon-dxw:
[Add user-agent to APIClient](https://trello.com/c/3dirfIAY/1319-add-user-agent-to-apiclient)

tests/client/test_client.py
```diff
@@ -173,3 +174,10 @@ def test_format_uri_trailing_slash(self):
     def test_format_uri_all_the_slashes(self):
         uri = "/ewca/2022/123/"
         assert self.client._format_uri_for_marklogic(uri) == "/ewca/2022/123.xml"
+
+    def test_user_agent(self):
+        user_agent = self.client.session.prepare_request(
+            Request("GET", "http://example.invalid")
+        ).headers["user-agent"]
+        assert user_agent.startswith("ds-caselaw-marklogic-api-client/")
+        assert "." in user_agent  # i.e. looks like some sort of semantic version
```
anthonyhashemi, line 183: can we change it so it's just an assertion on a regex pattern?

I think:
`assert re.match(r'^ds-caselaw-marklogic-api-client\/\d+\.\d+\.\d+$', user_agent)`
Could be what we want?

Sticking
`^ds-caselaw-marklogic-api-client\/\d+\.\d+\.\d+$` into the pattern in https://regex101.com/
and testing for example `ds-caselaw-marklogic-api-client/2.3.1` 

works.

If we expect a v before the version number, then we can just add that in the pattern

anthonyhashemi (approved):

PR #364, opened 2023-08-30
Be more strict about the flavours of strings

jacksonj04:
There is currently ambiguity about the types of strings which are being passed around, particularly with the overloaded term "URI". This uses `typing.NewType` to create some new types of `str` so we can be explicit in what we mean. These are:

- `Document.DocumentURIString`: Our generally accepted concept of a document's unique identifier
- `xquery_type_dicts.MarkLogicDocumentURIString`: A URI referring to a concrete document within MarkLogic
- `xquery_type_dicts.MarkLogicDocumentVersionURIString`: A subclass of the above, referring to a specific version of a document in MarkLogic
- `xquery_type_dicts.MarkLogicPrivilegeURIString`: A URI referring to a user privilege within MarkLogic.

To allow for this, the `build_xquery_type_dicts` script will interpret some variable names as always meaning a specific type of URI, either document, version or privilege as above. These are then reflected in the typed dicts.

### DIFF
--- a/script/build_xquery_type_dicts
+++ b/script/build_xquery_type_dicts
@@ -12,8 +12,12 @@ These classes are automatically generated using the `script/build_xquery_type_di
 checks. They are used to enforce appropriately typed variables being passed in to MarkLogic XQuery functions.
 \"\"\"
 
-from typing import Any, Optional, TypedDict
+from typing import Any, NewType, Optional, TypedDict
 
+MarkLogicDocumentURIString = NewType("MarkLogicDocumentURIString", str)
+MarkLogicDocumentVersionURIString = NewType("MarkLogicDocumentVersionURIString", MarkLogicDocumentURIString)
+
+MarkLogicPrivilegeURIString = NewType("MarkLogicPrivilegeURIString", str)
 
 class MarkLogicAPIDict(TypedDict):
     pass"""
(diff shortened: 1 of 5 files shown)


dragon-dxw:
[only read description] <3

script/build_xquery_type_dicts
```diff
@@ -12,8 +12,12 @@ These classes are automatically generated using the `script/build_xquery_type_di
 checks. They are used to enforce appropriately typed variables being passed in to MarkLogic XQuery functions.
 \"\"\"
 
-from typing import Any, Optional, TypedDict
+from typing import Any, NewType, Optional, TypedDict
 
+MarkLogicDocumentURIString = NewType("MarkLogicDocumentURIString", str)
```
dragon-dxw, line 17: We should probably defined these once somewhere.

src/caselawclient/xquery_type_dicts.py
```diff
@@ -6,72 +6,76 @@
 checks. They are used to enforce appropriately typed variables being passed in to MarkLogic XQuery functions.
 """
 
-from typing import Any, Optional, TypedDict
+from typing import Any, NewType, Optional, TypedDict
 
+MarkLogicDocumentURIString = NewType("MarkLogicDocumentURIString", str)
```
dragon-dxw, line 11: Is this the right place for these definitions to live, or do we want a apiclient.typing module?

dragon-dxw (approved):

src/caselawclient/xquery_type_dicts.py
jacksonj04: It's not my favourite place for them – at first they lived in `Client` but this led to some circular import pain during testing, so I just moved them closer to where they're actually used.

We could bikeshed about this forever, and I'm fairly confident that a more clear place for things like this to exist will become apparent as we start doing more of properly nailing down the last few remaining bits of spit-and-sticky-tape stuff.

PR #371, opened 2023-09-01
Move mypy-boto to main requirements, update requirements

dragon-dxw:
Mypy-boto-s3 and -sns are imported in non-test files, so should be installed as a part of the custom API client in all cases.

Also update all requirements.


anthonyhashemi (approved):

dragon-dxw:
Anthony pointed out another option from https://mypy-boto3.readthedocs.io/en/latest/

### Pylint compatibility

It is totally safe to use TYPE_CHECKING flag in order to avoid boto3-stubs dependency in production. However, there is an issue in pylint that it complains about undefined variables. To fix it, set all types to object in non-TYPE_CHECKING mode.
import boto3
from typing import TYPE_CHECKING

```
if TYPE_CHECKING:
    from mypy_boto3.ec2 import EC2Client, EC2ServiceResource
    from mypy_boto3.ec2.waiters import BundleTaskCompleteWaiter
    from mypy_boto3.ec2.paginators import DescribeVolumesPaginator
else:
    EC2Client = object
    EC2ServiceResource = object
    BundleTaskCompleteWaiter = object
    DescribeVolumesPaginator = object
```

dragon-dxw:
[💻 Mypy imports in APIclient should be prod, not dev dependencies](https://trello.com/c/KyDXbaO2/1324-%F0%9F%92%BB-mypy-imports-in-apiclient-should-be-prod-not-dev-dependencies)

PR #376, opened 2023-09-06
Access most recent transformation/enrichment dates

dragon-dxw:
https://trello.com/c/ArC1muhC/1321-api-client-expose-date-time-metadata-for-each-version-for-version-page-changes-1306

Yields the most recent date of each type as a datetime.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ src/caselawclient/__pycache__/
 __pycache__
 __pypackages__
 .coverage
+.gradle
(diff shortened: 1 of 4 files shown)


jacksonj04 (commented):
Super duper picky; could we turn these into two functions each:

- Get list of dates of type _foo_
- Get the latest date of type _foo_

It just means we can be more granular with the testing, and separates the logic (slightly) of the XPath querying and the deciding which one of them is right.

timcowlishaw (approved):

jacksonj04 (approved):
Looks shiny!

PR #377, opened 2023-09-06
Change `is_failure` behaviour to rely on `failed_to_parse`

jacksonj04:
The `is_failure` parameter previously relied on checking the URI to determine if a failure state existed. This is a hangover from previous behaviour, but not one we want.

Change the behaviour so that it instead relies on checking other parameters which are indicative of a failure. At the moment this is only `failed_to_parse`, but this list may be expanded in the future.

### DIFF
--- a/src/caselawclient/models/documents.py
+++ b/src/caselawclient/models/documents.py
@@ -70,11 +70,6 @@ class Document:
             "This document failed to parse",
         ),
         (
-            "is_failure",
-            False,
-            "This {document_noun} does not have a valid URI",
-        ),
-        (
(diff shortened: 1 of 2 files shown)


dragon-dxw:
This currently controls whether a delete button is available.

https://github.com/nationalarchives/ds-caselaw-editor-ui/blob/b81c195721ac744bb8ae18d8b6cc6ca7a120336d/ds_caselaw_editor_ui/templates/includes/judgment/toolbar.html#L47

dragon-dxw (approved):

PR #379, opened 2023-09-07
Provide versions of a document as a list of Documents

dragon-dxw:


timcowlishaw:
Thanks so much for this @dragon-dxw ! Agree with @jacksonj04's suggestion but otherwise it looks ideal.

anthonyhashemi (approved):

PR #380, opened 2023-09-07
Add forgotten tests for helper

dragon-dxw:

### DIFF
--- a/tests/client_helpers/test_xml_helpers.py
+++ b/tests/client_helpers/test_xml_helpers.py
@@ -1,0 +1,15 @@
+import lxml.etree
+
+from caselawclient.xml_helpers import get_xpath_match_string, get_xpath_match_strings
+
+
+def test_xpath_single():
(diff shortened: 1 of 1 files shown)


tests/client_helpers/test_xml_helpers.py
```diff
@@ -0,0 +1,15 @@
+import lxml.etree
+
+from caselawclient.xml_helpers import get_xpath_match_string, get_xpath_match_strings
+
+
+def test_xpath_single():
+    node = lxml.etree.fromstring("<root><cat>1</cat><cat>2</cat></root>")
+    path = "//cat/text()"
+    assert get_xpath_match_string(node, path) == "1"
```
jacksonj04, line 9: Happy with this PR as stands, note for future that we should probably rename this to `get_xpath_first_matching_string` or similar, _or_ have this method error out if it has more than one match

jacksonj04 (approved):

PR #391, opened 2023-09-15
Get annotation for a single document

dragon-dxw:


dragon-dxw:
Alternatively we could go down a route like:
```xquery
xquery version "1.0-ml";

import module namespace dls="http://marklogic.com/xdmp/dls" at "/MarkLogic/dls.xqy";

let $uri := "/eat/2022/1.xml"
let $versions := ($uri, dls:document-version-uris($uri))

for $version in $versions
    return <version><uri>{$version}</uri><annotation>{xdmp:document-properties($version)//dls:annotation/text()}</annotation></version>
```    
which returns a list of versions (including the current judgment) and their annotations.

smoketest/smoketest-setup.xqy
```diff
@@ -0,0 +1,37 @@
...
+          <uk:number>1234</uk:number>
+          <uk:cite>[2023] EAT 1234</uk:cite>
+          <uk:parser>0.13.3</uk:parser>
+          <uk:hash>1234</uk:hash>
+        </proprietary>
+    </meta>
+  </judgment>
+</akomaNtoso>
+
+(: return dls:document-delete($URI, fn:false(), fn:false()) :)
```
anthonyhashemi, line 36: can this commented line be deleted?

tests/client/test_get_annotation.py
```diff
@@ -0,0 +1,27 @@
+import unittest
+from unittest.mock import patch
+
+from caselawclient.Client import MarklogicApiClient
+
+
+class TestGetSetAnnotation(unittest.TestCase):
```
anthonyhashemi, line 7:
```suggestion
class TestGetAnnotation(unittest.TestCase):
```

anthonyhashemi: or are we adding a test for setting annotations in this class also?

anthonyhashemi (approved):
Couple of minor tidy up comments in the tests but looks good 👍

PR #395, opened 2023-09-18
Provide publication information when getting metadata for search results

dragon-dxw:
image

image

This is potentially breaking behaviour since there is a new `published` value editor_status can emit or `new` for Document.status.
https://linear.app/tna/issue/FCL-82/status-in-search-results-doesnt-match-the-actual-status-of-the

But it turns out that the editor just displays this value, never really parsing it; seems to be true for the Document too.

Editor could do with a colour tweak for "new"

### DIFF
--- a/tests/models/test_documents.py
+++ b/tests/models/test_documents.py
@@ -13,6 +13,7 @@ from caselawclient.errors import (
 from caselawclient.models.documents import (
     DOCUMENT_STATUS_HOLD,
     DOCUMENT_STATUS_IN_PROGRESS,
+    DOCUMENT_STATUS_NEW,
     DOCUMENT_STATUS_PUBLISHED,
     CannotPublishUnpublishableDocument,
     Document,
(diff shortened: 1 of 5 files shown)


tests/models/test_documents.py
```diff
@@ -155,19 +156,28 @@ def test_judgment_content_as_xml_tree(self, mock_api_client):
...
         on_hold_document = Document("test/1234", mock_api_client)
         on_hold_document.is_held = True
         on_hold_document.is_published = False
+        on_hold_document.assigned_to = "duck"
         assert on_hold_document.status == DOCUMENT_STATUS_HOLD
 
         published_document = Document("test/1234", mock_api_client)
         published_document.is_held = False
         published_document.is_published = True
+        published_document.assigned_to = "duck"
```
anthonyhashemi, line 180: quack

anthonyhashemi (approved):
Document.status and SearchResult.editor_status are now both in line, looks good 😎

tests/models/test_documents.py
dragon-dxw: 🦆

PR #402, opened 2023-09-26
Expose how the latest version came into being

timcowlishaw:
## Changes in this PR:

a new `document.get_latest_manifestation_type` method which returns either "tna-enrich" or "transform" depending on the "name" property of the most recent manifestation date.

This is used in the versions table in the editor UI to annotate each version with it's 'version type', which helps editors better understand where each version came from.

https://trello.com/c/jyTZaNMi/1359-eui-version-history-add-column-to-table-to-indicate-type-of-version

### DIFF
--- a/src/caselawclient/models/documents.py
+++ b/src/caselawclient/models/documents.py
@@ -216,6 +216,16 @@ class Document:
         else:
             return max(events)
 
+    def get_latest_manifestation_type(self) -> Optional[str]:
+        return max(
+            (
+                (type, time)
+                for type in ["transform", "tna-enriched"]
+                if (time := self.get_latest_manifestation_datetime(type))
+            ),
+            key=lambda x: x[1],
+        )[0]
+
     @cached_property
     def transformation_datetime(self) -> Optional[datetime.datetime]:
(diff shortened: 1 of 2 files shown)


src/caselawclient/models/documents.py
```diff
@@ -216,6 +216,14 @@ def get_latest_manifestation_datetime(
         else:
             return max(events)
 
+    def get_latest_manifestation_type(self) -> str:
+        if self.get_latest_manifestation_datetime(
+            "transform"
+        ) > self.get_latest_manifestation_datetime("tna-enriched"):
```
dragon-dxw, line 222: You might need to check if it's valid to compare None to a date time -- you might be better using the string version.

timcowlishaw: gave it a little refactor which hopefully makes things clearer!

timcowlishaw:
[[EUI] VERSION HISTORY Add column to table to indicate type of version](https://trello.com/c/jyTZaNMi/1359-eui-version-history-add-column-to-table-to-indicate-type-of-version)

dragon-dxw (approved):

PR #404, opened 2023-09-28
v15.1.2

dragon-dxw:
Skipping v15.1.1 because I accidently tried to make a tag with that.


timcowlishaw:
Oh, sorry, dragon, I just noticed this was open, and merged the new 'type of latest manifestation' branch! Does that mean that it'd make it into this release once it's merged, given that it's now on `main`? if so should i do a PR updating the changelog?

PR #411, opened 2023-10-03
Better version annotations

jacksonj04:
This PR does the following:

- [x] Allows annotations to be added to the first version of a document
- [ ] Renames `save_judgment_xml` to `update_document_xml`, so it's clearer what it's actually doing
- [ ] Makes annotations obligatory for all behaviours which create a new version
- [ ] Structures annotations so that we can extract richer data from them in future


timcowlishaw (commented):
Aha this all looks great to me. Only one small comment - are there sensible defaults we could provide in each context (insert_document_xml / update_document_xml etc) or is it better that we leave it up to the caller to be explicit? Happy either way, just thought it was worth pondering.

timcowlishaw (approved):

src/caselawclient/client_helpers/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+    """A class holding structured data about the reason for a version."""
+
+    def __init__(self, version_type: VersionType, message: Optional[str] = None):
+        """
+        :param version_type: The type of version being created
+        :param message: A human-readable string containing information about the version which can't be expressed in the
+            structured data.
+        """
+        self.version_type = version_type
+        self.message = message
```
timcowlishaw, line 37: oops, tiny thing Nick which i've just noticed when trying to use this in anger - we should probably set self.calling_function = None here or we end up with an AttributeError later when trying to use the annotation. In practice, this shouldn't happen, but it leads to some complications in tests

PR #420, opened 2023-10-12
poetry update

dragon-dxw:


jacksonj04 (approved):
We can get Renovate to automatically do these lockfile updates if we want (ie pushing to the most recent version matching the constraints).

PR #423, opened 2023-10-13
Add `automated` and `payload` to `VersionAnnotation`

jacksonj04:
`automated` is a required boolean, indicating if the version is the result of an automated process or a human action.

`payload` is an optional structured dict of additional information, which may be parsed by other tools such as EUI to expose relevant details.

### DIFF
--- a/src/caselawclient/client_helpers/__init__.py
+++ b/src/caselawclient/client_helpers/__init__.py
@@ -1,6 +1,6 @@
 import json
 from enum import Enum
-from typing import Optional, TypedDict
+from typing import Any, Optional, TypedDict
 
 from typing_extensions import NotRequired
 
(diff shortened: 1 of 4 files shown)


src/caselawclient/client_helpers/__init__.py
```diff
@@ -27,14 +29,25 @@ class VersionType(Enum):
 class VersionAnnotation:
     """A class holding structured data about the reason for a version."""
 
-    def __init__(self, version_type: VersionType, message: Optional[str] = None):
+    def __init__(
+        self,
+        version_type: VersionType,
```
dragon-dxw, line 34: Appreciate it's not new code, but importing two things from client_helper (VersionAnnotation and VersionType) to craft an annotation feels like it's breaking a boundary. Should we expose these in client directly?

dragon-dxw (approved):

PR #434, opened 2023-10-24
Highlight  query matches in document html

timcowlishaw:
## Changes in this PR:
 `document.content_as_html` now takes an optional `query` parameter which is preprocessed in the same manner as the search query, then used to highlight matching terms in the document, which are returned with a surrounding `<mark>` tag. This mark tag has an id which gives the number, in sequence of the match from 0..n in the form `mark_{x}`.

A new `document.number_of_mentions` method which takes a `query` parameter and returns the number of highlighted mentions returned in the html for the query

https://trello.com/c/tnpX6eJW/1412-search-highlight-term-within-a-judgment-feature-flag-test

### DIFF
--- a/src/caselawclient/models/documents.py
+++ b/src/caselawclient/models/documents.py
@@ -6,12 +6,14 @@ from typing import TYPE_CHECKING, Any, Dict, NewType, Optional
 from ds_caselaw_utils import courts
 from ds_caselaw_utils.courts import CourtNotFoundException
 from lxml import etree
+from lxml import html as html_parser
 from requests_toolbelt.multipart import decoder
 
 from caselawclient.models.utilities import extract_version
 
 from ..errors import (
     DocumentNotFoundError,
+    GatewayTimeoutError,
     NotSupportedOnVersion,
     OnlySupportedOnVersion,
 )
@@ -29,6 +31,10 @@ from .utilities.aws import (
 
 
 class UnparsableDate(Warning):
+    pass
+
+
+class GatewayTimeoutGettingHTMLWithQuery(RuntimeWarning):
     pass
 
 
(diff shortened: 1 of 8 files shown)


src/caselawclient/xquery/xslt_transform.xqy
```diff
@@ -14,6 +17,26 @@ let $xsl_path := fn:concat("judgments/xslts/", $xsl_filename)
...
+      <xsl:copy>
+        <xsl:apply-templates select="@*|node()"/>
+      </xsl:copy>
+    </xsl:template>
+    <xsl:template match="mark">
+      <xsl:copy>
+          <xsl:copy-of select="@*" />
+          <xsl:attribute name="id">
+              <xsl:text>mark_</xsl:text>
+              <xsl:value-of select="count(preceding::mark)"/>
```
dragon-dxw, line 33: There's a part of me that's worried that this might be accidently quadratic -- if there's a thousand marks, does this take a half-million searches?

it's probably fine?

dragon-dxw (approved):

src/caselawclient/models/documents.py
```diff
@@ -328,9 +328,13 @@ def content_as_xml_bytestring(self) -> bytes:
...
 
-    def content_as_html(self, version_uri: Optional[DocumentURIString] = None) -> str:
+    def content_as_html(
+        self,
+        version_uri: Optional[DocumentURIString] = None,
+        query: Optional[str] = None,
+    ) -> str:
         results = self.api_client.eval_xslt(
-            self.uri, version_uri, show_unpublished=True
+            self.uri, version_uri, show_unpublished=True, query=query
```
dragon-dxw, line 337: Slightly concerned about this being used for denial of service since it won't be cached, but actually what our backend does is pretty irrelevant to that compared to just random query parameters that don't hit anything.

src/caselawclient/xquery/xslt_transform.xqy
timcowlishaw: As discussed on slack, it doesn't appear to be O(n^2) and is actually fairly performant on big judgments - we have however added a check to guard against gateway timeouts and return the unhighlighted judgment, logging a warning, so we can monitor.

src/caselawclient/models/documents.py
timcowlishaw: I *think* my response to the previous question also covers this? let's test properly in staging and keep an eye out.

timcowlishaw:
Updated to address performance concerns and add a new `document.number_of_mentions` method

timcowlishaw:
Making this a draft momentarily - `cts:highlight` seems to break the markup that marklogic gives us back in some quite subtle and interesting ways. I'll debug properly before merging.

timcowlishaw:
Jim, being a genius, solved my weird markup problem with one line of xslt. I think this is ready!

PR #468, opened 2023-11-27
Refactor search to get rid of dependency on static `api_client` instance

jacksonj04:
Searching currently relies on the static instance of `api_client` which is initialised at `Client` load-time to interface with the database and retrieve metadata.

This instance is deprecated and due for removal, so we need to fix this behaviour. Search should only ever use the `Client` instance it was called with, so we need to either inject the dependency down or refactor the behaviour so it can directly access the `Client`.

As part of this, we move factory methods out of `SearchResultMetadata` and `SearchResult`, which provides a cleaner separation of concerns and reduces the depth to which we need to make changes to account for removing `api_client`.

There should be no functional changes as a result of this PR, and it should be non-breaking for EUI and PUI.

### DIFF
--- a/src/caselawclient/client_helpers/search_helpers.py
+++ b/src/caselawclient/client_helpers/search_helpers.py
@@ -1,3 +1,5 @@
+from lxml import etree
+
 from caselawclient.Client import MarklogicApiClient
 from caselawclient.responses.search_response import SearchResponse
 from caselawclient.search_parameters import SearchParameters
@@ -14,8 +16,10 @@ def search_judgments_and_parse_response(
 
     :return: The parsed search response as a SearchResponse object
     """
-    return SearchResponse.from_response_string(
-        api_client.search_judgments_and_decode_response(search_parameters)
+    return SearchResponse(
+        etree.fromstring(
+            api_client.search_judgments_and_decode_response(search_parameters)
+        )
     )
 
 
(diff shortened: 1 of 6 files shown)


src/caselawclient/client_helpers/search_helpers.py
```diff
@@ -14,8 +16,11 @@ def search_judgments_and_parse_response(
 
     :return: The parsed search response as a SearchResponse object
     """
-    return SearchResponse.from_response_string(
-        api_client.search_judgments_and_decode_response(search_parameters)
+    return SearchResponse(
```
timcowlishaw, line 19: Hmm, I feel like this could go further - could the the `Client` have  one public `search` method which takes `search_parameters`, and returns a `SearchResult`, handling the `search_judgments_and_decode_response` logic, parsing the element tree, and injecting itself as a dependency into the result? That way `search_helpers.py` would not need to know or care about the internals.

timcowlishaw (approved):
Looks like a great improvement to me, just one comment where we could perhaps refactor further! Not a blocker though.

src/caselawclient/client_helpers/search_helpers.py
jacksonj04: This has been added to the list of further refactoring.

PR #473, opened 2023-11-28
Clean up parser error handling

jacksonj04:
**Review and merge https://github.com/nationalarchives/ds-caselaw-custom-api-client/pull/472 first!**

Remove some unnecessary abstraction in how we identify if a document has failed to parse. At the same time, stop swallowing where a document is not XML (which should never happen) and fail hard with a specific exception instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@ The format is based on [Keep a Changelog 1.0.0].
 
 ## [Unreleased]
 
-- **Breaking:**: Fully remove the deprecated `caselawclient.api_client` instance.
+- **Breaking:** Fully remove the deprecated `caselawclient.api_client` instance.
+- **Feature:** New `Document.xml_root_element` function to replace `get_judgment_root`
+- **Feature:** Documents which are not valid XML are now identified by the raising of a new `Document.NonXMLDocumentError` exception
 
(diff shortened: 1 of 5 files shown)


tests/responses/test_search_result.py
```diff
@@ -14,32 +15,46 @@
...
+            patch.object(
+                self.client, "get_properties_for_search_results"
+            ) as mock_get_properties_for_search_results,
+            patch.object(self.client, "get_last_modified") as mock_get_last_modified,
+        ):
+            mock_get_properties_for_search_results.return_value = "<foo/>"
+            mock_get_last_modified.return_value = "bar"
+
+            node = etree.fromstring(valid_search_result_xml)
+            search_result = SearchResult(node, self.client)
```
dragon-dxw, line 43: This feels suspiciously like SearchResult should be a method on the client.

dragon-dxw (approved):

dragon-dxw (approved):

tests/responses/test_search_result.py
jacksonj04: I definitely agree, I've made a note in the list of things to refactor that search (as a whole) is in the wrong place. It's a bit of a chunk to move though and has a lot of test changes involved, so it wants to happen in its own PR.

PR #482, opened 2023-12-04
⚠️  Make parameter to set_published non-optional

dragon-dxw:
Technically a breaking change to the API but stops `api_client.set_published()` being a footgun that does nothing.


dragon-dxw:
Wait until the ingester doesn't do this before deploying.

jacksonj04:
@dragon-dxw this is good to be rebased and launched 🚀

PR #498, opened 2023-12-20
Get a list of documents waiting for enrichment

jacksonj04:
MarkLogic uses TDE to extract relevant fields and pop them into an SQL database, which we then query to get a report on documents waiting to be enriched.

### DIFF
--- a/src/caselawclient/xquery/get_pending_enrichment_for_version.xqy
+++ b/src/caselawclient/xquery/get_pending_enrichment_for_version.xqy
@@ -1,0 +1,18 @@
+xquery version "1.0-ml";
+
+declare variable $target_version as xs:int external;
+
+xdmp:to-json(xdmp:sql(
+  "SELECT process_data.uri, enrich_version_string, minutes_since_enrichment_request
+  FROM (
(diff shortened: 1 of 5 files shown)


src/caselawclient/xquery/get_pending_enrichment_for_version.xqy
```diff
@@ -0,0 +1,18 @@
...
+
+xdmp:to-json(xdmp:sql(
+  "SELECT process_data.uri, enrich_version_string, minutes_since_enrichment_request
+  FROM (
+    SELECT process_data.uri, enrich_version_string, enrich_major_version, DATEDIFF('minute', last_sent_to_enrichment, CURRENT_TIMESTAMP) AS minutes_since_enrichment_request
+    FROM documents.process_data
+    JOIN documents.process_property_data ON process_data.uri = process_property_data.uri
+  )
+  WHERE ((enrich_version_string IS NULL) OR (enrich_major_version < @target_version))
+  AND (minutes_since_enrichment_request > 1440 OR minutes_since_enrichment_request IS NULL)
```
dragon-dxw, line 13: worried that this might be a lot of reprocessing if something fails -- could we start at a month maybe?

dragon-dxw (approved):

PR #525, opened 2024-01-10
Return the highest enrichment version seen on MarkLogic

dragon-dxw:
Does not necessarily reflect the version number in Enrichment. This might lead to scenarios where bulk enrichment is stalled until a single successful manual or new-document enrichment occurs.

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -952,6 +952,21 @@ class MarklogicApiClient:
 
         return results
 
+    def get_highest_enrichment_version(self) -> int:
+        """This gets the highest enrichment version in the database,
+        so if nothing has been enriched with the most recent version of enrichment,
+        this won't reflect that change."""
+        table = json.loads(
+            get_single_string_from_marklogic_response(
+                self._send_to_eval(
+                    {},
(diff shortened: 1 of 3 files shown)


src/caselawclient/Client.py
```diff
@@ -952,6 +952,21 @@ def get_combined_stats_table(self) -> list[list[Any]]:
...
 
+    def get_highest_enrichment_version(self) -> int:
+        """This gets the highest enrichment version in the database,
+        so if nothing has been enriched with the most recent version of enrichment,
+        this won't reflect that change."""
+        table = json.loads(
+            get_single_string_from_marklogic_response(
+                self._send_to_eval(
+                    {},
+                    "get_highest_enrichment_version.xqy",
```
jacksonj04, line 963: I think this is missing!

jacksonj04 (changes requested):
Looks like missing a file in the commit.

src/caselawclient/Client.py
dragon-dxw: Added! Good spot.

jacksonj04 (approved):

PR #526, opened 2024-01-10
Prioritise quoted phrases in search snippets.

timcowlishaw:
## Changes in this PR:
This goes with a [marklogic PR](https://github.com/nationalarchives/ds-caselaw-marklogic/pull/50) which ensures that snippets containing quoted phrases in the query are given priority when they're displayed to users.

This PR provides the list of quoted phrases to marklogic (essentially impossible to do with marklogic's own regex engine), and also limits the number of matched phrases returned to 4 (the current default), as the number returned by marklogic itself has necessarily increased (we need to generate a larger amount of matches before prioritisation to ensure we catch any mentions of the quoted phrase).

https://trello.com/c/IQK3v7c5/1560-search-matching-text-should-reflect-search-query-in-quotations-over-other-words-in-search-results

### DIFF
--- a/src/caselawclient/search_parameters.py
+++ b/src/caselawclient/search_parameters.py
@@ -1,7 +1,9 @@
+import re
 from dataclasses import dataclass
 from typing import Any, Dict, List, Optional, Set
 
 RESULTS_PER_PAGE = 10
+QUOTED_PHRASE_REGEX = '"([^"]*)"'
 
 
 @dataclass
@@ -43,6 +45,7 @@ class SearchParameters:
             "show_unpublished": str(self.show_unpublished).lower(),
             "only_unpublished": str(self.only_unpublished).lower(),
             "collections": self._marklogic_collections,
+            "quoted_phrases": self._quoted_phrases,
         }
 
     @property
@@ -57,6 +60,12 @@ class SearchParameters:
         courts = self._court_list_splitter(court_text)
         alternative_court_names = self._get_alternative_court_names(courts)
         return list(courts | alternative_court_names)
+
+    @property
+    def _quoted_phrases(self) -> List[str]:
+        if self.query is None:
+            return []
+        return re.findall(QUOTED_PHRASE_REGEX, self.query)
 
     @staticmethod
     def _join_court_text(court_text: str) -> str:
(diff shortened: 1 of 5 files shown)


timcowlishaw:
[SEARCH: matching text should reflect search query in quotations over other words in search results](https://trello.com/c/IQK3v7c5/1560-search-matching-text-should-reflect-search-query-in-quotations-over-other-words-in-search-results)

src/caselawclient/search_parameters.py
```diff
@@ -58,6 +61,14 @@ def _marklogic_courts(self) -> Optional[List[str]]:
         alternative_court_names = self._get_alternative_court_names(courts)
         return list(courts | alternative_court_names)
 
+    @property
+    def _quoted_phrases(self) -> List[str]:
```
dragon-dxw, line 65: This might be more legible as
```
if self.query is None:
    return []
return re.findall(...)
```

tests/client/test_search_parameters.py
```diff
@@ -85,3 +87,14 @@ def test_courts(self, test_input, expected):
         """
         search_parameters = SearchParameters(court=test_input)
         assert set(search_parameters.as_marklogic_payload()["court"]) == set(expected)
+
+    def test_quoted_phrases(self):
```
dragon-dxw, line 91: Can we test multiple quotes, like `"a b" c "d e"` as well please? Want to be sure `" c "` doesn't pop up

dragon-dxw (approved):

tests/client/test_search_parameters.py
```diff
@@ -85,3 +87,14 @@ def test_courts(self, test_input, expected):
...
+        """
+        GIVEN a SearchParameters with a `query` containing a quoted phrase
+        WHEN calling the as_marklogic_payload() method
+        THEN the as_marklogic_payload() method should return a list of the quote phrases in the query.
+        """
+        search_parameters = SearchParameters(query='"Reasonable adjustment" transport')
+        assert search_parameters.as_marklogic_payload()["quoted_phrases"] == [
```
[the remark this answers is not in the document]
timcowlishaw: aha no it's all good - it needs to still appear in the `query` too! (although maybe that should be made more clear)

tests/client/test_search_parameters.py
timcowlishaw: aha good plan - I did a bunch of those tests in ipython and really should have put them in here :-)

src/caselawclient/search_parameters.py
timcowlishaw: aha nice, will amend!

PR #536, opened 2024-01-25
Nested parser instructions

dragon-dxw:
Remove "published" as a thing that we send, we should check locally on import.
Reformat to nested format.


dragon-dxw:
Discussed making the individual entries mandatory -- so court always exists, but None means "we don't know, please tell us parser"

jacksonj04 (approved):

PR #543, opened 2024-01-30
Add methods to set/get jurisdiction of a document, for GRC subdivisions

timcowlishaw:
## Changes in this PR:

The General Regulatory Chamber has a number of Jurisdictions (or
'sub-tribunals' as we sometimes call them). These are different to the
separate Lists we see within eg the Chancery Division - Lists are an
administrative distinction, decided on by the court themselves, whereas
Jurisdictions are a statutory one, decided in law. Neither jurisdiction
nor list affects the NCN of the judgment, but they're stored differently
in the XML - Lists are represented in the court code in the uk:court
element, whereas judgments are represented with the uk:jurisdiction
element in addition to the uk:court.

This PR provides accessors for the jurisdiction data, both in isolation,
and combined with the court as a slash-separated string, following the
convention established in the courts metedata in ds-caselaw-utils.

This will allow us to display the jurisdiction in the PUI (and allow
editors to correct it in EUI) with minimal changes

I have also fixed a bug that was showing up in testing for me - the parse and enrich timestamps were an hour out, as i'm in a different timezone :-)

https://trello.com/c/JDoWjYsa/1602-medium-grcfollow-logic-for-the-courts-and-expose-grc-subdivisions-on-pui-prototype-first

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -446,6 +446,24 @@ class MarklogicApiClient:
 
         return self._send_to_eval(vars, "set_metadata_court.xqy")
 
+    def set_document_jurisdiction(
+        self, document_uri: DocumentURIString, content: str
+    ) -> requests.Response:
+        uri = self._format_uri_for_marklogic(document_uri)
+        vars: query_dicts.SetMetadataJurisdictionDict = {"uri": uri, "content": content}
+        return self._send_to_eval(vars, "set_metadata_jurisdiction.xqy")
+
+    def set_document_court_and_jurisdiction(
+        self, document_uri: DocumentURIString, content: str
+    ) -> requests.Response:
+        if "/" in content:
+            court, jurisdiction = re.split("\\s*/\\s*", content)
+            self.set_document_court(document_uri, court)
+            return self.set_document_jurisdiction(document_uri, jurisdiction)
+        else:
+            self.set_document_court(document_uri, content)
+            return self.set_document_jurisdiction(document_uri, "")
+
     def set_judgment_this_uri(
         self, judgment_uri: DocumentURIString
     ) -> requests.Response:
--- a/src/caselawclient/models/documents.py
+++ b/src/caselawclient/models/documents.py
@@ -199,6 +199,20 @@ class Document:
         )
 
     @cached_property
+    def jurisdiction(self) -> str:
+        return self.xml.get_xpath_match_string(
+            "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:jurisdiction/text()",
+            {
+                "uk": "https://caselaw.nationalarchives.gov.uk/akn",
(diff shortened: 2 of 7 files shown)


src/caselawclient/models/documents.py
```diff
@@ -464,9 +478,9 @@ def enrich(self) -> None:
         """
         Request enrichment of the document
         """
-
+        now = datetime.datetime.now(datetime.timezone.utc)
```
dragon-dxw, line 481: I think this is going to be a subtle bug, because some datetimes coming back to the api client are going to be in naive and some will be datetime aware. We need to handle and test both cases.

dragon-dxw (changes requested):

src/caselawclient/models/documents.py
timcowlishaw: Hey dragon! aha, i've split that change out to another branch - will open a seperate PR for it. This one should now only have the changes related to the GRC subdivisions.

timcowlishaw:
[{Medium} GRC|Follow logic for the courts and expose GRC subdivisions on PUI - (prototype first)](https://trello.com/c/JDoWjYsa/1602-medium-grcfollow-logic-for-the-courts-and-expose-grc-subdivisions-on-pui-prototype-first)

jacksonj04 (approved):

tests/client/test_get_set_metadata.py
```diff
@@ -64,6 +64,62 @@ def test_set_document_court(self):
...
+        with patch.object(self.client, "eval") as mock_eval:
+            uri = "judgment/uri"
+            content = "court"
+            expected_vars = {"uri": "/judgment/uri.xml", "content": content}
+            self.client.set_document_court_and_jurisdiction(uri, content)
+
+            assert len(mock_eval.call_args_list) == 1
+
+            assert mock_eval.call_args.args[0] == (
+                os.path.join(ROOT_DIR, "xquery", "set_metadata_court.xqy")
```
dragon-dxw, line 119: This should probably explicitly delete the jurisdiction as it is not a part of the entered value.

src/caselawclient/Client.py
```diff
@@ -446,6 +446,23 @@ def set_document_court(
...
+    ) -> requests.Response:
+        uri = self._format_uri_for_marklogic(document_uri)
+        vars: query_dicts.SetMetadataJurisdictionDict = {"uri": uri, "content": content}
+        return self._send_to_eval(vars, "set_metadata_jurisdiction.xqy")
+
+    def set_document_court_and_jurisdiction(
+        self, document_uri: DocumentURIString, content: str
+    ) -> requests.Response:
+        if "/" in content:
+            court, jurisdiction = content.split("/")
```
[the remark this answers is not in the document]
timcowlishaw: Aha I've changed this to a regex which also matches whitespace to be sure :-)

jacksonj04 (approved):

PR #550, opened 2024-01-31
Expose jurisdictions in search results

timcowlishaw:
## Summary of changes

A couple of oversights in the GRC Jurisdictions work - Marklogic now exposes the jurisdiction in search results so we need to use it there to return the correct court.

I've also added some tests to make sure court validation works as expected with jurisdictions.

## Checklist


src/caselawclient/models/documents.py
```diff
@@ -210,7 +210,10 @@ def jurisdiction(self) -> str:
 
     @property
     def court_and_jurisdiction(self) -> str:
```
jacksonj04, line 212: After sleeping on it, could we perhaps have something like
```suggestion
    def court_and_jurisdiction_identifier_string(self) -> str:
```
and a docstring explaining where a list of the valid codes can be found? That way it's clear that this is actually a coded representation, rather than a naive string for users.

jacksonj04: Actually a followup, is it worth creating an explicit new type for this with `NewType` like we do for `DocumentURIString`, maybe `CourtIdentifierString` so we can always be really explicit about what we're expecting to pass around?

dragon-dxw (approved):

jacksonj04 (approved):

timcowlishaw:
[{Medium} GRC|Follow logic for the courts and expose GRC subdivisions on PUI - (prototype first)](https://trello.com/c/JDoWjYsa/1602-medium-grcfollow-logic-for-the-courts-and-expose-grc-subdivisions-on-pui-prototype-first)

PR #555, opened 2024-02-12
Add a helper to get press summaries for a document

pezholio:
## Summary of changes

This adds a `get_press_summaries_for_document_uri` which gets press summaries for a document URI. I’ve made the query itself configurable so we can use it for other components too, but for now we’re only using it for press summaries.

As I wasn’t super confident it was going to work, I’ve also added a smoketest, which I think it worth having around.

## Checklist

### DIFF
--- a/src/caselawclient/xquery/get_components_for_document.xqy
+++ b/src/caselawclient/xquery/get_components_for_document.xqy
@@ -1,0 +1,20 @@
+xquery version "1.0-ml";
+
+declare namespace akn = "http://docs.oasis-open.org/legaldocml/ns/akn/3.0";
+declare namespace uk = "https://caselaw.nationalarchives.gov.uk/akn";
+
+declare variable $parent_uri as xs:string external;
+declare variable $component as xs:string external;
+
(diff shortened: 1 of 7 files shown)


src/caselawclient/xquery/get_components_for_document.xqy
```diff
@@ -0,0 +1,12 @@
+xquery version "1.0-ml";
+
+declare namespace akn = "http://docs.oasis-open.org/legaldocml/ns/akn/3.0";
+
+declare variable $parent_uri as xs:string external;
+declare variable $component as xs:string external;
+
+let $matcher := concat($parent_uri, "/", $component ,"/*.xml")
```
jacksonj04, line 8: I wouldn't necessarily rely on this behaviour; at the moment press summaries all have a URI which begins with the URI of the parent document, but this is a particularly brittle bit of design. I'd be more inclined to look for documents with the right `<meta><neutral_citation>` (I think) value.

pezholio: Good call, I've made a tweak to the query which makes it more robust (I think!)

jacksonj04 (approved):

PR #557, opened 2024-02-13
Only enrich if not recently called

dragon-dxw:
Don't enrich if very recently sent to enrichment.

### DIFF
--- a/src/caselawclient/models/documents.py
+++ b/src/caselawclient/models/documents.py
@@ -31,6 +31,8 @@ from .utilities.aws import (
     uri_for_s3,
 )
 
+MINIMUM_ENRICHMENT_TIME = datetime.timedelta(minutes=20)
+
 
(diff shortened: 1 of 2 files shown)


pezholio (commented):
One suggestion that's based on some very loosely held reckons

src/caselawclient/models/documents.py
```diff
@@ -495,6 +497,14 @@ def enrich(self) -> None:
             enrich=True,
         )
 
+    def enrich(self, cooldown: datetime.timedelta = MINIMUM_REPARSE_TIME) -> bool:
```
pezholio, line 500: Is this method always triggered by a user interaction? I'm wondering if we might be worth having a `can_enrich?` method, which will ghost out / hide the enrich button if it returns false

pezholio (approved):
LGTM 👍

PR #558, opened 2024-02-14
(1695) Don't enrich when the document does not validate against a schema

pezholio:
This adds a `validates_against_schema` property to a document, and adds the logic to `can_enrich` to return true if a document was not recently enriched AND it passes schema validation.

### DIFF
--- a/tests/client/test_validate_document.py
+++ b/tests/client/test_validate_document.py
@@ -1,0 +1,62 @@
+import unittest
+from unittest.mock import patch
+
+from caselawclient.Client import MarklogicApiClient
+
+
+class TestValidateDocument(unittest.TestCase):
+    def setUp(self):
+        self.client = MarklogicApiClient("", "", "", False)
+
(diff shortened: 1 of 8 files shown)


tests/client/test_validate_document.py
```diff
@@ -0,0 +1,62 @@
...
+            assert self.client.validate_document("/foo/bar/123") is True
+
+    def test_validation_failure(self):
+        with patch.object(self.client, "eval") as mock_eval:
+            mock_eval.return_value.text = "true"
+            mock_eval.return_value.headers = {
+                "content-type": "multipart/mixed; boundary=c878f7cb55370005"
+            }
+            mock_eval.return_value.content = (
+                b"\r\n--c878f7cb55370005\r\n"
```
dragon-dxw, line 35: Is there a more sensible shorter version of this content that'd demonstrate the concept equally well?

dragon-dxw (approved):

PR #561, opened 2024-02-16
Check s3 file exists

dragon-dxw:
## Summary of changes

Be able to check if an S3 docx exists and logic for whether we allow reparse

## Checklist


pezholio (approved):
This looks good - it's a shame we don't have a class for all our S3-related methods, so we could mock out `create_s3_client`. This seems like a good workaround for now though 👍

pyproject.toml
```diff
@@ -26,6 +26,7 @@ boto3 = "^1.26.112"
 typing-extensions = "^4.7.1"
 mypy-boto3-s3 = "^1.26.104"
 mypy-boto3-sns = "^1.26.69"
+moto = {version = "^5.0.1", extras = ["all"]}
```
dragon-dxw, line 29: should probably be a dev dependency

src/caselawclient/models/documents.py
```diff
@@ -278,6 +284,10 @@ def transformation_datetime(self) -> Optional[datetime.datetime]:
     def enrichment_datetime(self) -> Optional[datetime.datetime]:
         return self.get_latest_manifestation_datetime("tna-enriched")
 
+    @cached_property
+    def reparse_datetime(self) -> Optional[datetime.datetime]:
+        return self.get_latest_manifestation_datetime("last_sent_to_parser")
```
dragon-dxw, line 289: This can't work, can it? This needs to be a MarkLogic property I think, rather than getting the thing from the XML document.

src/caselawclient/models/documents.py
```diff
@@ -627,6 +637,32 @@ def reparse(self) -> None:
...
+    def reparse(self) -> bool:
+        if self.can_reparse:
+            self.force_reparse()
+            return True
+        return False
+
+    @cached_property
+    def can_reparse(self) -> bool:
+        """
+        Is it sensible to enrich this document?
```
dragon-dxw, line 649: comment: change to reparse

src/caselawclient/models/utilities/aws.py
```diff
@@ -78,6 +78,20 @@ def generate_signed_asset_url(key: str) -> str:
...
+    bucket = env("PRIVATE_ASSET_BUCKET", None)
+    s3_key = generate_docx_key(uri)
+    client = create_s3_client()
+    try:
+        client.head_object(Bucket=bucket, Key=s3_key)
+        return True
+    except botocore.exceptions.ClientError as e:
+        if e.response["Error"]["Code"] == "404":
+            return False
+        raise
```
dragon-dxw, line 92: remove this raise

src/caselawclient/models/documents.py
dragon-dxw: The enrichment one can't work either -- it has to be looking at the properties!

CristinaRO (approved):
Approved in principle, based on the understanding that @dragon-dxw will address the things about which they left notes to self. :)

src/caselawclient/models/documents.py
```diff
@@ -518,9 +544,9 @@ def can_enrich(self) -> bool:
     @cached_property
     def enriched_recently(self) -> bool:
```
dragon-dxw, line 545: Function name potentially not accurate to functionality.

src/caselawclient/models/documents.py
```diff
@@ -627,6 +653,32 @@ def reparse(self) -> None:
...
+    def can_reparse(self) -> bool:
+        """
+        Is it sensible to reparse this document?
+        """
+        if (self.reparsed_recently is False) and self.docx_exists():
+            return True
+        return False
+
+    @cached_property
+    def reparsed_recently(self) -> bool:
```
dragon-dxw, line 672: function name potentially not accurate to functionality

tests/models/test_documents.py
```diff
@@ -427,28 +427,51 @@ def test_unpublish(
...
             tz=datetime.timezone.utc
         ) - datetime.timedelta(days=2)
 
         assert document.enriched_recently is False
 
+    def test_reparsed_recently_returns_false_when_never_enriched(self, mock_api_client):
+        document = Document("test/1234", mock_api_client)
+        mock_api_client.get_property.return_value = ""
+
+        assert document.enriched_recently is False
```
dragon-dxw, line 457: Incorrect assertion

dragon-dxw:
Stripped out the time-related portions as an unnecessary distraction.

PR #574, opened 2024-03-26
Fix/remove UUID prefix

dragon-dxw:
## Summary of changes

It [was commented](https://my.slack.com/archives/C02TP2L2Z0F/p1711375372312379) that our execution IDs were prefixed with `fcl_ex_id_` -- we can remove that.


adam-murray (approved):
LGTM!

PR #605, opened 2024-06-19
Make enrichment date maths not timezone aware

dragon-dxw:
https://app.rollbar.com/a/dxw/fix/item/tna-caselaw-editor-ui/163?utm_campaign=reactivated_item_message&utm_medium=slack&utm_source=rollbar-notification

New function to ensure we always return a zulu time.


dragon-dxw:
This is not a datetime we control, like when we save the answer to "what time did we send this to be enriched?", but an in-XML attribute which just comes back from enrichment. 

These appear to all be raw datetimes, no timezone; it uses now() which is datetime aware, so local time, but then doesn't output the timezone data. This is a bug, but it's not one we can fix here.

The question of *which* timezone that date is in is a valid question, but our date maths only needs to be accurate within days, and we're running it in the UK so it'll only be out by daylight saving.

dragon-dxw:
We radically overhauled the approach here -- now we make everything UTC all the time.

src/caselawclient/models/documents.py
```diff
@@ -251,7 +253,8 @@ def get_manifestation_datetimes(
             f"/akn:FRBRdate{name_filter}/@date",
             {"akn": "http://docs.oasis-open.org/legaldocml/ns/akn/3.0"},
         )
-        return [datetime.datetime.fromisoformat(event) for event in iso_datetimes]
+
+        return [parse_string_date(event, pytz.UTC) for event in iso_datetimes]
```
dragon-dxw, line 257: Is this the right place to define the timezone -- can we guarantee other XML non-specified times are UTC?

I'm pretty sure all the other ones are specified, and would be UTC if they're not.

jacksonj04: I'm going to go with yes, because we're only sure about it in this context (and even then it's only a 90% certainty)

src/caselawclient/models/utilities/dates.py
```diff
@@ -0,0 +1,20 @@
+from datetime import datetime
+
+from dateutil.parser import isoparse
+from pytz import UTC, tzinfo
+
+
+def parse_string_date(iso_string: str, timezone: tzinfo.BaseTzInfo) -> datetime:
```
jacksonj04, line 7: Is it worth naming this something like `parse_date_string_and_add_timezone`? Not totally sold on that exact wording, just a bit more explicit.

jacksonj04 (approved):

jacksonj04:
Some test failures to address in here as well.

jacksonj04 (changes requested):
Some tests are failing because of ISO string formatting, but the substance of the code looks fine 👍🏻

src/caselawclient/models/utilities/dates.py
dragon-dxw: gone with parse_date_string_as_utc

jacksonj04 (approved):

PR #607, opened 2024-06-26
Always set last sent to parser

dragon-dxw:
Whilst changing the editor to send at least one document with the crontab, discovered that the process will fail because if a document is rejected (e.g. doesn't have a docx) it never gets the "I tried to send" message.

(We did never send this to the parser, though...)

"We should rename all occurances of last_sent_to_parser" is a fair comment; this would involve changing sql_properties_extract in marklogic and get_pending_parse_for_version.xqy in this repo.

### DIFF
--- a/src/caselawclient/models/documents.py
+++ b/src/caselawclient/models/documents.py
@@ -635,6 +635,10 @@ class Document:
         )
 
     def reparse(self) -> bool:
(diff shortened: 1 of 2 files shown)


jacksonj04 (changes requested):
Can we get a quick test that this is always called?

jacksonj04 (approved):

PR #661, opened 2024-08-28
[FCL-268] Add functionality for interacting with document identifiers

jacksonj04:


dragon-dxw (approved):

src/caselawclient/models/documents/__init__.py
```diff
@@ -70,7 +63,7 @@ class Document:
 
     attributes_to_validate: list[tuple[str, bool, str]] = [
         (
-            "failed_to_parse",
+            "is_failure",
```
dragon-dxw, line 66: Reasoning about this change is hard: this feels like potentially a major breaking change and struggling to review it.

tests/models/documents/test_document_verbs.py
```diff
@@ -226,10 +222,16 @@ def test_force_reparse_empty(self, sns, mock_api_client):
...
+
+        document.neutral_citation = "[2023] Test 123"
+        document.consignment_reference = "TDR-12345"
+
+        document.body.name = "Judgment v Judgement"
+        document.body.court = "Court of Testing"
+        document.body.document_date_as_date = datetime.date(2023, 2, 3)
 
-        Judgment.force_reparse(document)
+        document.force_reparse()
```
dragon-dxw, line 234: Ah; I think the reason I was doing it the other way was that the mock document wasn't actually a judgment. Having an actual document is much cleaner.

src/caselawclient/models/documents/__init__.py
jacksonj04: `failed_to_parse` is now implemented in `DocumentBody`, not `Body`. `is_failure` currently proxies this and only this, but was left as an abstraction because it could potentially be expanded in future to include other things we consider to be a total failure state.

By making the `Document` validity checks care only about this abstraction method it's less likely that future changes to what we consider a failure would be missed.

PR #666, opened 2024-08-29
[FCL-217] Enable ruff on API Client

jacksonj04:
## Summary of changes

Enable ruff on the API Client, run obvious auto-fixes, and enable some bare-bones rules to help improve code quality.

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,6 +19,8 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.6.3
     hooks:
+      - id: ruff
+        args: [--fix, --exit-non-zero-on-fix]
       - id: ruff-format
 
   - repo: local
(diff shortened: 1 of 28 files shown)


.pre-commit-config.yaml
```diff
@@ -19,6 +19,8 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.6.3
     hooks:
+      - id: ruff
+        args: [--fix, --exit-non-zero-on-fix]
```
dragon-dxw, line 23: Does exit-non-zero-on-fix correctly commit the fixes?

jacksonj04: Yes - we want an exit code of zero _only_ if there have been no changes, otherwise the non-zero code prompts pre-commit that things weren't ok and that something in the commit needs changing (even if that's just adding the things which `--fix` has tidied up for us).

dragon-dxw (approved):

PR #695, opened 2024-09-26
[FCL-331] Improve test factory behaviour

jacksonj04:
## Summary of changes

Test factories now live within the package proper (so they can be used by downstream consumers for consistent test behaviour), and now also provide real instances of the Document class instead of a mock. This gives more parity in behaviour, as well as reducing mock boilerplate.

## Jira

FCL-331


src/caselawclient/factories.py
```diff
@@ -0,0 +1,168 @@
...
+            xml_string = kwargs.pop("xml")
+        else:
+            xml_string = "<akomantoso>This is some XML of a judgment.</akomantoso>"
+
+        document_body = DocumentBody(
+            xml_bytestring=xml_string.encode(encoding="utf-8"),
+        )
+
+        for param_name, default_value in cls.PARAMS_MAP.items():
+            if param_name in kwargs:
```
dragon-dxw, line 35: is `kwargs.get(param_name, default_value)` cleaner?

src/caselawclient/factories.py
```diff
@@ -0,0 +1,168 @@
...
+        "source_email": "uploader@example.com",
+        "consignment_reference": "TDR-12345",
+        "assigned_to": "",
+        "versions": [],
+    }
+
+    target_class: TypeAlias = Document
+
+    @classmethod
+    def build(cls, uri: str = "test/2023/123", **kwargs: Any) -> target_class:
```
dragon-dxw, line 61: Does it make sense to have the explicitly named kwargs (html, body) as explicit optional parameters?

dragon-dxw (approved):

src/caselawclient/Client.py
```diff
@@ -800,10 +800,7 @@ def eval_xslt(
             else None
         )
 
-        if os.getenv("XSLT_IMAGE_LOCATION"):
-            image_location = os.getenv("XSLT_IMAGE_LOCATION")
-        else:
-            image_location = ""
+        image_location = os.getenv("XSLT_IMAGE_LOCATION") if os.getenv("XSLT_IMAGE_LOCATION") else ""
```
dragon-dxw, line 803: recommend `os.getenv("XSLT_IMAGE_LOCATION", "")`

src/caselawclient/models/neutral_citation_mixin.py
```diff
@@ -40,14 +40,8 @@ def neutral_citation(self) -> str:
...
-        return True
+        return bool(self.neutral_citation)
 
     @cached_property
     def has_valid_ncn(self) -> bool:
-        if not self.has_ncn or not neutral_url(self.neutral_citation):
-            return False
-
-        return True
+        return self.has_ncn and neutral_url(self.neutral_citation) is not None
```
dragon-dxw, line 47: I'm not convinced this is correct. 
Neutral citation is a string, so this should probably be `!= ""`.

This brings into doubt a pile of other simplifications...

src/caselawclient/models/utilities/dates.py
```diff
@@ -9,11 +9,7 @@ def parse_string_date_as_utc(iso_string: str, timezone: tzinfo.BaseTzInfo) -> da
     ensure that it is converted to a UTC-aware datetime"""
 
     mixed_date = isoparse(iso_string)
-    if not mixed_date.tzinfo:
-        # it is an unaware time
-        aware_date = timezone.localize(mixed_date)
-    else:
-        aware_date = mixed_date
+    aware_date = timezone.localize(mixed_date) if not mixed_date.tzinfo else mixed_date
```
dragon-dxw, line 12: I'd be tempted to invert this -- `... = mixed_date if mixed_date.tzinfo else timezone.localize(...)`

dragon-dxw (changes requested):
I think there's at least one significant bug in the auto simplification.

dragon-dxw (approved):

PR #712, opened 2024-10-08
Fix typing issues

dragon-dxw:
## Summary of changes

Fix some typing which causes linting to fail.

## Checklist


src/caselawclient/models/judgments.py
```diff
@@ -23,17 +25,19 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+                {
+                    "uk": "https://caselaw.nationalarchives.gov.uk/akn",
+                    "akn": "http://docs.oasis-open.org/legaldocml/ns/akn/3.0",
+                },
+            )
         )
 
     @property
-    def best_human_identifier(self) -> str:
+    def best_human_identifier(self) -> NeutralCitationString:
```
jacksonj04, line 40: This should still be a `str`; the fact it's an NCN right now doesn't mean it won't be something less specific in future.
```suggestion
    def best_human_identifier(self) -> str:
```

src/caselawclient/models/press_summaries.py
```diff
@@ -25,16 +27,18 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+                "/akn:akomaNtoso/akn:doc/akn:preface/akn:p/akn:neutralCitation/text()",
+                {
+                    "akn": "http://docs.oasis-open.org/legaldocml/ns/akn/3.0",
+                },
+            )
         )
 
     @property
-    def best_human_identifier(self) -> str:
+    def best_human_identifier(self) -> NeutralCitationString:
```
jacksonj04, line 41: This should still be a `str`; the fact it's an NCN right now doesn't mean it won't be something less specific in future.
```suggestion
    def best_human_identifier(self) -> str:
```

src/caselawclient/models/utilities/move.py
```diff
@@ -30,7 +31,7 @@ def update_document_uri(
     :param api_client: An instance of MarklogicApiClient used to make the search request
     :return: The URL associated with the `target_citation`
     """
-    new_ncn_based_uri = caselawutils.neutral_url(target_citation.strip())
+    new_ncn_based_uri = caselawutils.neutral_url(NeutralCitationString(target_citation.strip()))
```
jacksonj04, line 34: I think we could push this up into the method signature, so `target_citation` has to be a `NeutralCitationString`. That way it'll force stuff downstream to also use that type.

jacksonj04 (changes requested):

src/caselawclient/models/utilities/move.py
jacksonj04: See some changes in https://github.com/nationalarchives/ds-caselaw-custom-api-client/pull/705/commits/0876d66894885658e86b2f0528d418ff11d269a7 if you want to just lift them – my only "don't do this" would be typing `api_client`, because it forces us to fix a load of annotation stuff.

Which we should, but I don't want to hold everything else up for it.

src/caselawclient/models/judgments.py
jacksonj04: I think this one is still missed!

jacksonj04 (approved):

PR #716, opened 2024-10-09
Lock judgment with fixed timeout

dragon-dxw:
## Summary of changes

In order to support automatic unlocking of enriched judgments before the editors start work in the morning, allow a fixed amount of time before automatically unlocking the judgment.

Once used in the privileged API, this should prevent enrichment leaving editors with locked judgments when enrichment fails.

## Checklist

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -634,12 +634,14 @@ class MarklogicApiClient:
         judgment_uri: DocumentURIString,
         annotation: str = "",
         expires_at_midnight: bool = False,
+        timeout_seconds: int = -1,
     ) -> requests.Response:
+        """If timeout_seconds is -1, the lock never times out"""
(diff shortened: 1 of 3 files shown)


src/caselawclient/Client.py
```diff
@@ -634,12 +634,14 @@ def checkout_judgment(
         judgment_uri: DocumentURIString,
         annotation: str = "",
         expires_at_midnight: bool = False,
+        timeout_seconds: int = -1,
```
jacksonj04, line 637: My only thought would be if an infinite lock is the right default behaviour, but I guess it's easy enough to fix if we discover the answer is "no".

jacksonj04 (approved):

src/caselawclient/Client.py
dragon-dxw: That's fair; it is the existing default behaviour.

PR #732, opened 2024-10-21
[FCL-386] Add native XSLT transformations to the API

jacksonj04:
This will allow us to run XSLT operations in the application itself, rather than handing it off to MarkLogic. This makes the process more easily testable and observable.

## Jira

FCL-386

### DIFF
--- a/src/caselawclient/models/documents/body.py
+++ b/src/caselawclient/models/documents/body.py
@@ -1,10 +1,12 @@
 import datetime
+import os
 import warnings
 from functools import cached_property
 from typing import Optional
 
 import pytz
 from ds_caselaw_utils.types import CourtCode
+from saxonche import PySaxonProcessor
 
 from caselawclient.models.utilities.dates import parse_string_date_as_utc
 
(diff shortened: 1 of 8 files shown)


src/caselawclient/models/documents/body.py
```diff
@@ -129,6 +131,23 @@ def enrichment_datetime(self) -> Optional[datetime.datetime]:
     def content_as_xml(self) -> str:
         return self._xml.xml_as_string
 
+    @cache
+    def content_as_html(self, image_base_url: Optional[str] = None) -> str:
+        """Convert the XML representation of the Document into HTML for rendering."""
+
+        html_xslt_location = os.path.join(os.path.dirname(os.path.realpath(__file__)), "transforms", "html.xsl")
```
dragon-dxw, line 138: This can/will get out of sync with the MarkLogic XSLT which will be a cause of confusing issues.

tests/models/documents/test_document_body.py
```diff
@@ -272,3 +274,23 @@ def test_document_with_no_dates(self, mock_api_client):
         assert body.transformation_datetime is None
         assert body.get_latest_manifestation_datetime() is None
         assert body.get_manifestation_datetimes("any") == []
+
+    def test_content_as_html_for_standard_judgment(self):
+        """Run our HTML XSLT on a judgment to make sure it's formatting as we expect."""
+
+        with open(
+            os.path.join(os.path.dirname(os.path.realpath(__file__)), "xslt", "test_standard_judgment.xml"), "r"
```
dragon-dxw, line 282: these are surprisingly horrid

src/caselawclient/models/documents/body.py
jacksonj04: The plan is for MarkLogic to no longer handle XSLT once this work is completed, so hopefully the overlap period will be short!

PR #751, opened 2024-10-28
[FCL-461] Initialising a Document now requires a DocumentURIString, not a str

jacksonj04:
This previously accepted a str and would attempt to strip it and remove slashes. It now only accepts a `DocumentURIString` object.

To enforce this, `DocumentURIString` is now a subclass of `str` which will validate incoming strings on creation time and raise an exception if they don't match the expected format.

BREAKING CHANGE: Document can now no longer be initialised with a string as the `uri`, it must be a `DocumentURIString`.

## Checklist

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -209,12 +209,14 @@ class MarklogicApiClient:
         Returns a list of PressSummary objects associated with a given Document URI
         """
         vars: query_dicts.GetComponentsForDocumentDict = {
-            "parent_uri": DocumentURIString(uri if uri.startswith("/") else "/" + uri),
+            "parent_uri": uri,
             "component": "pressSummary",
         }
         response = self._send_to_eval(vars, "get_components_for_document.xqy")
         uris = get_multipart_strings_from_marklogic_response(response)
-        return [PressSummary(uri.strip(".xml"), self) for uri in uris]
+        return [
+            PressSummary(DocumentURIString(uri.strip("/").strip(".xml")), self) for uri in uris
+        ]  # TODO: Migrate this strip behaviour into proper manipulation of a MarkLogicURIString
 
     def get_document_by_uri(
         self,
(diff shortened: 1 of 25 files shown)


dragon-dxw (dismissed):

dragon-dxw (changes requested):
Can the initialisation function for DocumentURIString ensure that there are no prefix/trailing slashes

src/caselawclient/Client.py
```diff
@@ -209,12 +209,14 @@ def get_press_summaries_for_document_uri(
...
         vars: query_dicts.GetComponentsForDocumentDict = {
-            "parent_uri": DocumentURIString(uri if uri.startswith("/") else "/" + uri),
+            "parent_uri": uri,
             "component": "pressSummary",
         }
         response = self._send_to_eval(vars, "get_components_for_document.xqy")
         uris = get_multipart_strings_from_marklogic_response(response)
-        return [PressSummary(uri.strip(".xml"), self) for uri in uris]
+        return [
+            PressSummary(DocumentURIString(uri.strip("/").strip(".xml")), self) for uri in uris
```
dragon-dxw, line 218: Is there scope for a "parse unclean string" method on DocumentURIString

dragon-dxw (approved):
We should probably add a MarkLogicURIString.as_Document_URI() and DocumentURIString.as_MarkLogicURIString() and use them, but that isn't urgent.

jacksonj04:
> We should probably add a MarkLogicURIString.as_Document_URI() and DocumentURIString.as_MarkLogicURIString() and use them, but that isn't urgent.

Strong agree, it's on the list!

src/caselawclient/Client.py
jacksonj04: Yes; I think it should be a helper function, and `MarkLogicDocumentURIString.as_DocumentURIString()` should basically just call that so we can easily flip between the two.

PR #755, opened 2024-10-30
require published before bulk enrichment

dragon-dxw:
## Summary of changes

I'm not sure this the behaviour we actually want, given things should now be getting unlocked better?

## Checklist

### DIFF
--- a/src/caselawclient/xquery/get_pending_enrichment_for_version.xqy
+++ b/src/caselawclient/xquery/get_pending_enrichment_for_version.xqy
@@ -11,20 +11,26 @@ xdmp:to-json(xdmp:sql(
   "SELECT process_data.uri, enrich_version_string, minutes_since_enrichment_request
   FROM (
     SELECT
+      propertysummary.published,
       process_data.uri,
       enrich_version_string, enrich_major_version, enrich_minor_version,
       parser_major_version, parser_minor_version,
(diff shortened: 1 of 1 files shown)


jacksonj04 (approved):

src/caselawclient/xquery/get_pending_enrichment_for_version.xqy
```diff
@@ -11,20 +11,26 @@ xdmp:to-json(xdmp:sql(
...
-  AND (minutes_since_enrichment_request > 43200 OR minutes_since_enrichment_request IS NULL)
+  WHERE 
+    (minutes_since_enrichment_request > 43200 OR minutes_since_enrichment_request IS NULL) AND
+    (propertysummary.published = 'true') AND
+    ( enrich_version_string IS NULL 
+      OR
+      (
+        (enrich_major_version <= @target_enrichment_major_version AND enrich_minor_version < @target_enrichment_minor_version)
+           AND 
+        (parser_major_version = @target_parser_major_version AND parser_minor_version = @target_parser_minor_version)
```
jacksonj04, line 31: I think these terms should be `>=`, to allow for the possibility there's been a parser update but where we haven't updated the target version.

jacksonj04 (changes requested):

jacksonj04 (approved):

PR #778, opened 2024-11-18
[FCL-309] Add the ability for documents to store abstract identifiers in MarkLogic properties

jacksonj04:
## Summary of changes

- [x] New `IdentifierSchema` and `Identifier` abstract classes which define interfaces for how we think about what any given type of identifier looks like (`IdentifierSchema`), and how we might be able to interact with specific instances of that identifier (`Identifier`)
- [x] New `NeutralCitationSchema` and `NeutralCitation` concretions of the above abstract classes, defining NCN-specific implementations of methods where necessary.
- [x] `Identifier`s can compile themselves down to a URL slug, used for simplifying retrieval at any given URL.
- [ ] New methods for a `Document` which allow CRUD operations on its identifiers, using the above classes to turn identifier concepts into serialisations which can be stashed in the MarkLogic properties for a document.

## Checklist

## Jira

FCL-309

### DIFF
--- a/src/caselawclient/models/documents/__init__.py
+++ b/src/caselawclient/models/documents/__init__.py
@@ -127,7 +127,7 @@ class Document:
     Individual document classes should extend this list where necessary to validate document type-specific attributes.
     """
 
-    identifiers: dict[str, Identifier] = {}
+    _identifiers: dict[str, Identifier]
 
     def __init__(self, uri: DocumentURIString, api_client: "MarklogicApiClient", search_query: Optional[str] = None):
         """
@@ -151,7 +151,7 @@ class Document:
         )
         """ `Document.body` represents the body of the document itself, without any information such as version tracking or properties. """
 
-        self.identifiers = {}
+        self._initialise_identifiers()
 
     def __repr__(self) -> str:
         name = self.body.name or "un-named"
@@ -167,29 +167,41 @@ class Document:
         """There is a docx in S3 private bucket for this Document"""
         return check_docx_exists(self.uri)
 
-    def _load_identifiers(self) -> None:
-        """Load this document's identifiers from MarkLogic"""
-        identifiers_element_as_string = self.api_client.get_property(self.uri, "identifiers")
-        identifiers_element_as_etree = etree.fromstring(identifiers_element_as_string)
+    def _initialise_identifiers(self) -> None:
+        """Load this document's identifiers from MarkLogic."""
 
-        for identifier_etree in identifiers_element_as_etree.findall("identifier"):
-            identifier = unpack_identifier_from_etree(identifier_etree)
-            self.add_identifier(identifier)
+        self._identifiers = {}
+
+        identifiers_element_as_etree = self.api_client.get_property_as_node(self.uri, "identifiers")
+
+        if identifiers_element_as_etree is not None:
+            for identifier_etree in identifiers_element_as_etree.findall("identifier"):
+                identifier = unpack_identifier_from_etree(identifier_etree)
+                self.add_identifier(identifier)
+
+    @property
+    def identifiers(self) -> list[Identifier]:
+        """Return a list of Identifier objects for easy display and interaction."""
+        return list(self._identifiers.values())
 
     def add_identifier(self, identifier: Identifier) -> None:
         """Add an identifier to this Document's identifiers array."""
-        self.identifiers[identifier.uuid] = identifier
 
+        self._identifiers[identifier.uuid] = identifier
+
+    @property
     def identifiers_as_etree(self) -> etree._Element:
         identifiers_root = etree.Element("identifiers")
 
-        for identifier in self.identifiers.values():
+        for identifier in self.identifiers:
             identifiers_root.append(identifier.as_xml_tree)
 
         return identifiers_root
 
     def save_identifiers(self) -> None:
         """Save the current state of this Document's identifiers to MarkLogic."""
+
+        self.api_client.set_property_as_node(self.uri, "identifiers", self.identifiers_as_etree)
 
     @property
     def best_human_identifier(self) -> Optional[str]:
--- a/src/caselawclient/models/documents/__init__.py
+++ b/src/caselawclient/models/documents/__init__.py
@@ -16,6 +16,7 @@ from caselawclient.errors import (
     NotSupportedOnVersion,
     OnlySupportedOnVersion,
 )
+from caselawclient.models.identifiers import Identifier
 from caselawclient.models.identifiers.unpacker import unpack_identifier_from_etree
 from caselawclient.models.utilities import VersionsDict, extract_version, render_versions
 from caselawclient.models.utilities.aws import (
@@ -31,7 +32,6 @@ from caselawclient.models.utilities.aws import (
     uri_for_s3,
 )
 
-from ..identifiers import Identifier
 from .body import DocumentBody
 from .exceptions import CannotPublishUnpublishableDocument, DocumentNotSafeForDeletion, InvalidDocumentURIException
 from .statuses import DOCUMENT_STATUS_HOLD, DOCUMENT_STATUS_IN_PROGRESS, DOCUMENT_STATUS_NEW, DOCUMENT_STATUS_PUBLISHED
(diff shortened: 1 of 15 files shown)


src/caselawclient/models/documents/__init__.py
```diff
@@ -29,6 +29,7 @@
     uri_for_s3,
 )
 
+from ..identifiers import Identifier
```
dragon-dxw, line 34: is caselawclient.models.identifiers clearer?

src/caselawclient/models/documents/__init__.py
```diff
@@ -160,6 +161,10 @@ def docx_exists(self) -> bool:
         """There is a docx in S3 private bucket for this Document"""
         return check_docx_exists(self.uri)
 
+    @property
+    def identifiers(self) -> list[Identifier]:
```
dragon-dxw: A docstring would be nice

src/caselawclient/models/identifiers/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+    value: str
+
+    def __init_subclass__(cls: type["Identifier"], **kwargs: Any) -> None:
+        for required in ("schema",):
+            if not getattr(cls, required, False):
+                raise NotImplementedError(f"Can't instantiate Identifier without {required} attribute.")
+        super().__init_subclass__(**kwargs)
+
+    def __repr__(self) -> str:
+        return "self.schema.name: self.value"
```
dragon-dxw: f"..."

src/caselawclient/models/identifiers/neutral_citation.py
```diff
@@ -0,0 +1,35 @@
+import re
+
+from . import Identifier, IdentifierSchema
+
+VALID_NCN_PATTERN = re.compile(r"(^\[([0-9]{4})\] ([a-zA-Z]+)(?: ([a-zA-Z]+))? ([0-9]+)(?: \(([a-zA-Z]+)\))?$)")
```
dragon-dxw, line 5: I think I'd like named capture groups

tests/models/identifiers/test_identifier_ncns.py
```diff
@@ -0,0 +1,53 @@
...
+    )
+    def test_ncn_schema_validation_passes(self, value):
+        schema = neutral_citation.NeutralCitationNumberSchema
+        assert schema.validate_identifier(value) is True
+
+    @pytest.mark.parametrize(
+        "value",
+        [
+            "",
+            "bananas" "1604] EWCA Crim 555",
```
dragon-dxw: I think there's a `,` missing here after bananas

src/caselawclient/models/identifiers/__init__.py
```diff
@@ -24,25 +33,46 @@ def __repr__(self) -> str:
...
 
 class Identifier(ABC):
-    """
-    A base class for subclasses representing a concrete identifier.
-    """
+    """A base class for subclasses representing a concrete identifier."""
 
     schema: type[IdentifierSchema]
+
+    uuid: str
```
dragon-dxw, line 45: An identifier has a UUID -- I'm not sure how this is useful; do we want a prefix to mean we can disambiguate an arbitary Identifier UUID from a document UUID?

src/caselawclient/models/identifiers/unpacker.py
```diff
@@ -0,0 +1,33 @@
+from lxml import etree
+
+from . import Identifier, InvalidIdentifierXMLRepresentationException
+from .neutral_citation import NeutralCitationNumber
+
+IDENTIFIER_NAMESPACE_MAP: dict[str, type[Identifier]] = {
```
dragon-dxw, line 6: A little worried that this is out of the way for the Big List of Namespaces -- maybe this wants to live somewhere more central?

tests/models/identifiers/test_identifiers.py
```diff
@@ -0,0 +1,38 @@
+from lxml import etree
+
+from caselawclient.models.identifiers import Identifier, IdentifierSchema
+
+
+class TestIdentifierSchema(IdentifierSchema):
+    __test__ = False
```
dragon-dxw, line 7: what is this?

src/caselawclient/models/documents/__init__.py
```diff
@@ -125,6 +127,8 @@ class Document:
     Individual document classes should extend this list where necessary to validate document type-specific attributes.
     """
 
+    identifiers: dict[str, Identifier] = {}
```
dragon-dxw, line 130: ⚠️  concerned that this might be the same `{}` each time.

dragon-dxw: And we're saying self.identifiers later anyway.

src/caselawclient/models/documents/__init__.py
```diff
@@ -161,10 +167,29 @@ def docx_exists(self) -> bool:
...
+        """Load this document's identifiers from MarkLogic"""
+        identifiers_element_as_string = self.api_client.get_property(self.uri, "identifiers")
+        identifiers_element_as_etree = etree.fromstring(identifiers_element_as_string)
+
+        for identifier_etree in identifiers_element_as_etree.findall("identifier"):
+            identifier = unpack_identifier_from_etree(identifier_etree)
+            self.add_identifier(identifier)
+
+    def add_identifier(self, identifier: Identifier) -> None:
+        """Add an identifier to this Document's identifiers array."""
```
dragon-dxw, line 180: not an array

src/caselawclient/models/documents/__init__.py
```diff
@@ -161,10 +167,29 @@ def docx_exists(self) -> bool:
...
+    def identifiers_as_etree(self) -> etree._Element:
+        identifiers_root = etree.Element("identifiers")
+
+        for identifier in self.identifiers.values():
+            identifiers_root.append(identifier.as_xml_tree)
+
+        return identifiers_root
+
+    def save_identifiers(self) -> None:
+        """Save the current state of this Document's identifiers to MarkLogic."""
```
dragon-dxw, line 192: TODO

dragon-dxw:
I'm a touch concerned about the properties being used because they might be stale if overwritten, but that's a risk anyway at the moment.

Do we want more general properties parsing and editing? (Probably not)

tests/models/identifiers/test_identifiers.py
jacksonj04: This indicates to pytest that although the class is named as `Test` it doesn't actually contain things which want to be tested (this is the "Test Identifier Schema", not tests _for_ the Identifier Schema).

src/caselawclient/models/identifiers/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+            if not getattr(cls, required, False):
+                raise NotImplementedError(f"Can't instantiate IdentifierSchema without {required} attribute.")
+        super().__init_subclass__(**kwargs)
+
+    def __repr__(self) -> str:
+        return self.name
+
+    @classmethod
+    @abstractmethod
+    def validate_identifier(cls, value: str) -> bool:
```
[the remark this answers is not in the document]
jacksonj04: Making it an `@abstractmethod` raises a runtime for us every time a subclass is instantiated without it being initialised.

src/caselawclient/models/identifiers/__init__.py
jacksonj04: Added an `id-` prefix, since this is basically zero cost.

src/caselawclient/models/identifiers/neutral_citation.py
jacksonj04: I don't disagree, but it's a bit of extra work we don't need at the moment. I've left some notes that when we move all the NCN-based URL generation code into this class that we should tidy this up with named groups.

src/caselawclient/models/identifiers/unpacker.py
jacksonj04: The problem is that if we put it too central then we start to get circular import problems, which is why it's out here.

dragon-dxw (approved):

PR #789, opened 2024-11-29
Fcl 309 Add CRUD functions to identifiers

dragon-dxw:
## Summary of changes

Add basic crud on Document.Identifiers

On top of https://github.com/nationalarchives/ds-caselaw-custom-api-client/pull/778

### DIFF
--- a/src/caselawclient/models/identifiers/__init__.py
+++ b/src/caselawclient/models/identifiers/__init__.py
@@ -127,6 +127,14 @@ class Identifiers(dict[str, Identifier]):
         else:
             super().__delitem__(key)
 
+    def delete_type(self, deleted_identifier_type: type[Identifier]) -> None:
+        "For when we want an identifier to be the only valid identifier of that type, delete the others first"
+        uuids = self.keys()
+        for uuid in list(uuids):
+            # we could use compare to .schema instead, which would have diffferent behaviour for subclasses
+            if isinstance(self[uuid], deleted_identifier_type):
+                del self[uuid]
+
     @property
     def as_etree(self) -> etree._Element:
         """Return an etree representation of all the Document's identifiers."""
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "ds-caselaw-marklogic-api-client"
-version = "28.0.0"
+version = "29.0.0-alpha.2"
```
jacksonj04, line 3: I think this is leftover from testing, we can remove this and then do a proper version bump later when SQIDs are in.

jacksonj04 (changes requested):
I think the version change is errant, and it needs a CHANGELOG update, but other than that this looks good! Nice tidying up into `Identifiers` as well, that'll really help with encapsulation.

pyproject.toml
dragon-dxw: Done.

src/caselawclient/models/identifiers/__init__.py
```diff
@@ -101,6 +101,10 @@ def as_xml_tree(self) -> etree._Element:
     def url_slug(self) -> str:
         return self.schema.compile_identifier_url_slug(self.value)
 
+    def similar(self, other: "Identifier") -> bool:
```
jacksonj04, line 104: Bit bikesheddy; would this be better as `same_as(other)`?

jacksonj04 (commented):
One little bikesheddy comment about function naming.

src/caselawclient/models/identifiers/__init__.py
```diff
@@ -127,6 +127,14 @@ def __delitem__(self, key: Union[Identifier, str]) -> None:
         else:
             super().__delitem__(key)
 
+    def delete_type(self, deleted_identifier_type: type[Identifier]) -> None:
```
jacksonj04, line 130: I'm happy this existing for now (because we're being a bit blunt-instrument about NCNs), but when we handle B-citations this should probably go away and be replaced by something giving deletions more precision and intent.

jacksonj04 (approved):

PR #792, opened 2024-12-03
[FCL-499, FCL-532] Add IdentifierSchema for creating SQID-based Find Case Law Identifiers

jacksonj04:
- Adds a new FCLID identifier class
- Adds new `Client` method for getting the next sequence number from MarkLogic
- Calling `Document.publish()` will now assign the document a FCLID if one doesn't already exist.

### DIFF
--- a/src/caselawclient/models/identifiers/fclid.py
+++ b/src/caselawclient/models/identifiers/fclid.py
@@ -1,0 +1,48 @@
+import re
+from typing import TYPE_CHECKING
+
+from sqids import Sqids
+
+from . import Identifier, IdentifierSchema
+
+if TYPE_CHECKING:
+    from caselawclient.Client import MarklogicApiClient
+
+
+VALID_FCLID_PATTERN = re.compile(r"^[bcdfghjkmnpqrstvwxyz23456789]{4,}$")
+
+FCLID_MINIMUM_LENGTH = 8
+FCLID_ALPHABET = "bcdfghjkmnpqrstvwxyz23456789"
+
+sqids = Sqids(
+    min_length=FCLID_MINIMUM_LENGTH,
+    alphabet=FCLID_ALPHABET,
+)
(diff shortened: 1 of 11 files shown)


src/caselawclient/models/identifiers/fclid.py
```diff
@@ -0,0 +1,48 @@
...
+from . import Identifier, IdentifierSchema
+
+if TYPE_CHECKING:
+    from caselawclient.Client import MarklogicApiClient
+
+
+VALID_FCLID_PATTERN = re.compile(r"^[bcdfghjkmnpqrstvwxyz23456789]{4,}$")
+
+FCLID_MINIMUM_LENGTH = 8
+FCLID_ALPHABET = "bcdfghjkmnpqrstvwxyz23456789"
```
dragon-dxw, line 15: The repeated string irks me... `r"^[" + alphabet + r"]{4,}$"` isn't cleaner, though, nor is `{alphabet}{{4,}}`...

tests/models/documents/test_document_verbs.py
```diff
@@ -14,6 +14,7 @@
     DocumentURIString,
 )
 from caselawclient.models.identifiers import Identifiers
+from caselawclient.models.identifiers.fclid import FindCaseLawIdentifier
```
dragon-dxw, line 17: Inconsistent that we have "fclid" and "neutral_citation" as identifier modules.

dragon-dxw (approved):
Wary that we only store the FCLID in the identifiers, which is in the metadata which we poorly understand the ways in which versioning interacts with it, and is not necessarily well preserved when restoring collisions. We don't have a convienient way of seeing historical assignments -- it might be worth considering making an append-only log with as much info as possible. (This might also wind up in the XML at a later date, which will be more resilient to adhoc copying)

PR #794, opened 2024-12-06
Resolve identifier URIs to MarkLogic URIs

dragon-dxw:
## Summary of changes
Functions to take a PUI url matching an identifier's URI and return the MarkLogic URI.
https://national-archives.atlassian.net/browse/FCL-496

## Checklist

### DIFF
--- a/src/caselawclient/identifier_resolution.py
+++ b/src/caselawclient/identifier_resolution.py
@@ -1,0 +1,43 @@
+import json
+from typing import NamedTuple
+
+from caselawclient.models.documents import DocumentURIString
+from caselawclient.xquery_type_dicts import MarkLogicDocumentURIString
+
+
+class IdentifierResolutions(list["IdentifierResolution"]):
+    """
+    A list of candidate MarkLogic documents which correspond to a Public UI uri
+
+    MarkLogic returns a list of dictionaries; IdentifierResolution handles a single dictionary
+    which corresponds to a single identifier to MarkLogic document mapping.
+
+    see `xquery/resolve_from_identifier.xqy` and `resolve_from_identifier` in `Client.py`
+    """
+
+    @staticmethod
+    def from_marklogic_output(table: list[str]) -> "IdentifierResolutions":
+        return IdentifierResolutions(list(IdentifierResolution.from_marklogic_output(row) for row in table))
+
+    def published(self) -> "IdentifierResolutions":
+        "Filter the list so that only published documents are returned"
+        return IdentifierResolutions(list(x for x in self if x.document_published))
+
+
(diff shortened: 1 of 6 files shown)


src/caselawclient/identifier_resolution.py
```diff
@@ -0,0 +1,31 @@
+import json
```
dragon-dxw, line 1: I'm a bit conflicted about where this file lives -- I think this is better than being in `models.identifiers`. Maybe it should be in `models.identifier_resolution` ?

jacksonj04: I agree this isn't a data model - current location seems fine to me.

src/caselawclient/identifier_resolution.py
```diff
@@ -0,0 +1,31 @@
...
+
+class IdentifierResolution(NamedTuple):
+    identifier_uuid: str
+    document_uri: MarkLogicDocumentURIString
+    identifier_slug: DocumentURIString
+    document_published: bool
+
+    @staticmethod
+    def from_marklogic_output(raw_row: str) -> "IdentifierResolution":
+        row = json.loads(raw_row)
```
jacksonj04, line 16: This would be a candidate for refactoring in future if we do much more with TDE; possibly one to touch on when we improve how EUI reports on things in various pending states, or when we do exception reporting.

src/caselawclient/identifier_resolution.py
```diff
@@ -0,0 +1,31 @@
+import json
+from typing import NamedTuple
+
+from caselawclient.models.documents import DocumentURIString
+from caselawclient.xquery_type_dicts import MarkLogicDocumentURIString
+
+
+class IdentifierResolution(NamedTuple):
```
jacksonj04, line 8: Can we pop some docstrings against this, just to explain what it's doing?

src/caselawclient/xquery/resolve_from_identifier.xqy
```diff
@@ -0,0 +1,17 @@
+xquery version "1.0-ml";
+
+declare namespace xdmp="http://marklogic.com/xdmp"; 
+declare variable $identifier_uri as xs:string external;
+declare variable $published_only as xs:int? external := 0;
```
jacksonj04, line 5: Is it safer for this to default to being true, so we have to explicitly request that it include unpublished stuff?

jacksonj04 (approved):
Needs a touch more in docstrings and I think we should flip a default to the 'safe' choice, otherwise LGTM!

src/caselawclient/identifier_resolution.py
dragon-dxw: Done

PR #801, opened 2024-12-10
Use save_identifiers on document not identifiers.save()

dragon-dxw:
## Summary of changes

Move saving identifiers to the document


jacksonj04 (approved):
This feels a lot more solid to me!

dragon-dxw:
Add a test --
given a particular Document with a known value of Identifiers (probably mock this out tbh)
when I call document.save_identifiers() it
calls identifiers.validate and
calls set_property_as_node with expected values (edited)

jacksonj04 (approved):

PR #810, opened 2024-12-18
[FCL-533] ⚠️ Introduce methods to select the best identifier in various cases

jacksonj04:
### BREAKING CHANGE

- Methods which were previously guaranteed to return a Neutral Citation may now return `None`.

### Feat

- **FCL-533**: getting scored or preferred identifiers can now be done by type
- **FCL-533**: modify human identifier to rely on identifiers framework
- **FCL-533**: add scoring to Identifiers

### Fix

- **IdentifierSchema**: use hasattr instead of getattr with a default when testing required attributes

### DIFF
--- a/src/caselawclient/models/documents/__init__.py
+++ b/src/caselawclient/models/documents/__init__.py
@@ -174,7 +174,7 @@ class Document:
     @property
     def best_human_identifier(self) -> Optional[Identifier]:
         """Return the preferred identifier for the document, providing that it is considered human readable."""
-        preferred_identifier = self.identifiers.preferred
+        preferred_identifier = self.identifiers.preferred()
         if preferred_identifier and preferred_identifier.schema.human_readable:
             return preferred_identifier
         return None
--- a/src/caselawclient/models/documents/__init__.py
+++ b/src/caselawclient/models/documents/__init__.py
@@ -15,6 +15,7 @@ from caselawclient.errors import (
     NotSupportedOnVersion,
     OnlySupportedOnVersion,
 )
+from caselawclient.models.identifiers import Identifier
 from caselawclient.models.identifiers.fclid import FindCaseLawIdentifier, FindCaseLawIdentifierSchema
 from caselawclient.models.identifiers.unpacker import unpack_all_identifiers_from_etree
 from caselawclient.models.utilities import VersionsDict, extract_version, render_versions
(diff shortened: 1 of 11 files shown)


src/caselawclient/models/documents/__init__.py
```diff
@@ -171,13 +172,11 @@ def _initialise_identifiers(self) -> None:
...
 
     @property
-    def best_human_identifier(self) -> Optional[str]:
-        """
-        Some identifier that is understood by legal professionals to refer to this legal event
-        that is not the name of the document.
-        Typically, this will be the neutral citation number, should it exist.
-        Should typically be overridden in subclasses.
-        """
+    def best_human_identifier(self) -> Optional[Identifier]:
```
dragon-dxw, line 175: Do all identifiers have a sensible `__str__` such that this is a good drop-in replacement despite the type change?

src/caselawclient/models/identifiers/__init__.py
```diff
@@ -32,13 +32,20 @@ class IdentifierSchema(ABC):
...
+
     def __init_subclass__(cls: type["IdentifierSchema"], **kwargs: Any) -> None:
         """Ensure that subclasses have the required attributes set."""
         for required in (
             "name",
             "namespace",
+            "human_readable",
         ):
-            if not getattr(cls, required, False):
+            if not hasattr(cls, required):
```
dragon-dxw, line 48: :+1:

src/caselawclient/models/identifiers/__init__.py
```diff
@@ -101,6 +108,11 @@ def as_xml_tree(self) -> etree._Element:
     def url_slug(self) -> str:
         return self.schema.compile_identifier_url_slug(self.value)
 
+    @property
```
dragon-dxw, line 111: This does mean that an underspecified identifier will score highly... no other default seems sensible, though.

dragon-dxw (approved):

dragon-dxw:
Could you check whether this causes any typing issues in EUI/PUI?

jacksonj04:
It definitely does cause typing problems, but they're expected (eg it's places where the apps assume documents _always_ have NCNs). I'm in the process of fixing in EUI (see #1812), and this will be a major version bump.

src/caselawclient/models/documents/__init__.py
jacksonj04: No, but I want this to explode where we're not using it properly. This is definitely a breaking change.

PR #814, opened 2024-12-19
Preferred identifier now correctly handles case where there are none of type

jacksonj04:


jacksonj04:
Closed in favour of hotfixing in #815.

PR #827, opened 2025-01-02
str(identifier) should be just the value

dragon-dxw:
## Summary of changes

Before:
image

Make str(identifier) be something that's printable in the website, rather than falling back to the repr

### DIFF
--- a/src/caselawclient/models/utilities/aws.py
+++ b/src/caselawclient/models/utilities/aws.py
@@ -251,8 +251,6 @@ def request_parse(
         },
     }
 
-    # breakpoint()
-
     client.publish(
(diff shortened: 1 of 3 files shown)


src/caselawclient/models/utilities/aws.py
```diff
@@ -251,8 +251,6 @@ def request_parse(
         },
     }
 
-    # breakpoint()
```
jacksonj04, line 254: Good delete! Wonder why ruff didn't catch this?

jacksonj04 (approved):
Seems sensible enough to me. 🚢

PR #834, opened 2025-01-07
[FCL-568] Add new class for Press Summary identifiers

jacksonj04:
This adds a new identifier schema for relating Press Summaries to judgments.

This isn't a perfect solution - this kind of thing would ideally be handled by proper document relations and resolution code - but it does solve the problem in the short term without us needing to also add a whole new system for handling relations.

### DIFF
--- a/src/caselawclient/models/identifiers/press_summary_ncn.py
+++ b/src/caselawclient/models/identifiers/press_summary_ncn.py
@@ -1,0 +1,20 @@
+from .neutral_citation import NeutralCitationNumber, NeutralCitationNumberSchema
+
+
+class PressSummaryRelatedNCNIdentifierSchema(NeutralCitationNumberSchema):
+    """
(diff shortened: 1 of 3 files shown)


src/caselawclient/models/identifiers/press_summary_ncn.py
```diff
@@ -0,0 +1,20 @@
+from .neutral_citation import NeutralCitationNumber, NeutralCitationNumberSchema
+
+
+class PressSummaryRelatedNCNIdentifierSchema(NeutralCitationNumberSchema):
+    """
+    Identifier schema for relating a Press Summary to a Judgment with a given NCN
+    """
+
+    name = "Press Summary relates to NCN"
+    namespace = "psrelatedncn"
```
dragon-dxw, line 10: Are we happy this is the right namespace name? others have been uk / fcl prefixed which felt like a category

dragon-dxw (approved):

PR #855, opened 2025-01-23
Ensure publishing documents only looks in s3 'folder'

dragon-dxw:
https://national-archives.atlassian.net/browse/FCL-617

We noticed that when publishing ewca/civ/2023/1 the editor was causing every document to be copied from the unpublished assets bucket to the published assets bucket that started with this prefix; this includes files such as ewca/civ/2023/1000/image1.jpg

## Checklist

### DIFF
--- a/src/caselawclient/models/utilities/aws.py
+++ b/src/caselawclient/models/utilities/aws.py
@@ -2,7 +2,7 @@ import datetime
 import json
 import logging
 import uuid
-from typing import Any, Literal, Optional, TypedDict, overload
+from typing import TYPE_CHECKING, Any, Literal, Optional, TypedDict, overload
 
 import boto3
 import botocore.client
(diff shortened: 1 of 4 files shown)


src/caselawclient/models/utilities/aws.py
```diff
@@ -58,8 +67,9 @@ def create_sns_client() -> SNSClient:
     return create_aws_client("sns")
 
 
-def uri_for_s3(uri: str) -> str:
-    return uri.lstrip("/")
+def uri_for_s3(uri: DocumentURIString) -> S3PrefixString:
```
jacksonj04, line 70:
```suggestion
def uri_for_s3(uri: DocumentURIString) -> S3PrefixedString:
```

src/caselawclient/models/utilities/aws.py
```diff
@@ -13,9 +13,18 @@
...
 
+if TYPE_CHECKING:
+    from caselawclient.models.documents import DocumentURIString
+else:
+    DocumentURIString = None
+
 env = environ.Env()
 
 
+class S3PrefixString(str):
```
jacksonj04, line 24: Is it worth adding some validation here (see eg https://github.com/nationalarchives/ds-caselaw-custom-api-client/blob/87799a998ddea62fe7f0c9fce12ba00f25b798c3/src/caselawclient/models/documents/__init__.py#L58) to prove that the string ends with a `/`?

jacksonj04 (approved):
Couple of tiny naming twiddles, but otherwise 👍🏻

src/caselawclient/models/utilities/aws.py
jacksonj04: Actually thinking about it this isn't prefixed at all, it's suffixed! Maybe something like `S3DocumentFolderString`? I know it's not _really_ a folder, but...

dragon-dxw: I think it is a "prefix" itself.

jacksonj04: Agreed, I was approaching this from the wrong angle.

jacksonj04 (approved):

PR #866, opened 2025-01-30
⚠️ document.body.content_as_html can be None

dragon-dxw:
## Summary of changes

If there is no parsed DOCX in the XML (i.e. there is no content in the header or body) `Document.body.content_as_html` now returns `None`.

This will cause typing changes downstream as it's now an `Optional[str]`

You can pre-emptively check with `Document.body.has_actual_content`

https://national-archives.atlassian.net/browse/FCL-640

## Checklist


jlhdxw (approved):
LGTM

PR #868, opened 2025-01-31
Relax has_content to not require valid akomaNtoso

dragon-dxw:
There were many errors on PUI caused by the examples having short XML which doesn't have `header` or `judgmentBody` tags. Let's be casual about what we expect to receive.


dragon-dxw:
Do Press Summaries use a different tag than judgmentBody?

dragon-dxw:
Rewritten to iterate through "tags after meta" to see if any contain or are followed by non-whitespace

PR #886, opened 2025-02-06
[FCL-667] Retrieve a DocumentURIString from a MarkLogic one

dragon-dxw:
## Summary of changes

We need to be able to switch both directions between a DocumentURIString (`uksc/2024/1`) and a MarklogicDocumentURIString (`/uksc/2024/1.xml`)

### DIFF
--- a/src/caselawclient/types.py
+++ b/src/caselawclient/types.py
@@ -2,8 +2,29 @@ class InvalidDocumentURIException(Exception):
     """The document URI is not valid."""
 
 
+class InvalidMarkLogicDocumentURIException(Exception):
+    """The MarkLogic document URI is not valid."""
+
+
 class MarkLogicDocumentURIString(str):
-    pass
+    def __new__(cls, content: str) -> "MarkLogicDocumentURIString":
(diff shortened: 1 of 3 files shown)


dragon-dxw:
Could do with some tests but was being a blocker.

jacksonj04 (changes requested):
Happy with the function, but we want to have a few test cases!

src/caselawclient/types.py
```diff
@@ -2,8 +2,29 @@ class InvalidDocumentURIException(Exception):
...
-    pass
+    def __new__(cls, content: str) -> "MarkLogicDocumentURIString":
+        # Check that the URI begins with a slash
+        if content[0] != "/":
+            raise InvalidMarkLogicDocumentURIException(
+                f'"{content}" is not a valid MarkLogic document URI; URIs must begin with a slash.'
+            )
+
+        # Check that the URI ends with ".xml"
+        if content[-4:] != ".xml":
```
dragon-dxw, line 18: `.endswith(".xml")`?

jacksonj04 (approved):

PR #894, opened 2025-02-11
Use short name for slug

dragon-dxw:
## Summary of changes
Add `.slug` to Document to avoid having to type `document.identifiers.preferred().uri_slug` (which isn't quite valid due to preferred returning None)

## Checklist


src/caselawclient/factories.py
```diff
@@ -70,13 +71,18 @@ def build(
...
+            return [document]
+
         if not api_client:
             api_client = Mock(spec=MarklogicApiClient)
             api_client.get_judgment_xml_bytestring.return_value = DEFAULT_DOCUMENT_BODY_XML.encode(encoding="utf-8")
             api_client.get_property_as_node.return_value = None
 
         document = cls.target_class(uri, api_client=api_client)
         document.body = kwargs.pop("body") if "body" in kwargs else DocumentBodyFactory.build()
+        document.identifiers.add(FindCaseLawIdentifier(value="a1b2c3"))
```
jacksonj04, line 84: This should really behave similarly to `body`, so maybe something like

```suggestion
if "identifiers" in kwargs:
    for identifier in kwargs.pop("identifiers"):
        document.identifiers.add(identifier)
else:
   
 document.identifiers.add(FindCaseLawIdentifier(value="a1b2c3"))
```

and make sure `identifier` is an `Optional[list[Identifier]]`?

src/caselawclient/models/documents/__init__.py
```diff
@@ -161,7 +161,15 @@ def public_uri(self) -> str:
         """
         :return: The absolute, public URI at which a copy of this document can be found
         """
-        return f"https://caselaw.nationalarchives.gov.uk/{self.uri}"
+        return f"https://caselaw.nationalarchives.gov.uk/{self.slug}"
+
+    @cached_property
+    def slug(self) -> str:
```
jacksonj04, line 167: Needs a docstring, since what this is doing isn't immediately obvious.

src/caselawclient/models/documents/__init__.py
```diff
@@ -161,7 +161,15 @@ def public_uri(self) -> str:
...
+        preferred_identifier = self.identifiers.preferred()
+        if preferred_identifier:
+            return preferred_identifier.url_slug
+        msg = f"No preferred identifier exists for {self.uri}"
+        raise RuntimeError(msg)
```
jacksonj04, line 172: This makes sense to me, we should never be trying to use this in a context where documents don't at least have an FCLID.

jacksonj04 (changes requested):

src/caselawclient/models/documents/__init__.py
```diff
@@ -161,7 +161,19 @@ def public_uri(self) -> str:
         """
         :return: The absolute, public URI at which a copy of this document can be found
         """
-        return f"https://caselaw.nationalarchives.gov.uk/{self.uri}"
+        return f"https://caselaw.nationalarchives.gov.uk/{self.slug}"
+
+    @cached_property
+    def slug(self) -> str:
+        """
+        :return: The best public-facing URI for the judgment, which is the slug
```
jacksonj04, line 169:
```suggestion
        :return: The best public-facing URL for the judgment, which is the slug
```

jacksonj04 (approved):

PR #904, opened 2025-02-19
Stop validating xml against schema before triggering enrichment

anthonyhashemi:
https://national-archives.atlassian.net/browse/FCL-697

## Summary of changes

- Stop validating xml against schema before triggering enrichment
- Add logging for enrich method for debugging

### DIFF
--- a/src/caselawclient/models/documents/__init__.py
+++ b/src/caselawclient/models/documents/__init__.py
@@ -357,7 +357,7 @@ class Document:
         """
         Is it sensible to enrich this document?
         """
-        return (self.enriched_recently is False) and self.validates_against_schema
+        return self.enriched_recently is False
 
     @cached_property
     def enriched_recently(self) -> bool:
--- a/src/caselawclient/models/documents/__init__.py
+++ b/src/caselawclient/models/documents/__init__.py
@@ -347,17 +347,12 @@ class Document:
         """
         Request enrichment of a document, if it's sensible to do so.
         """
-        if self.can_enrich:
+        if self.enriched_recently is False:
+            print("Enrichment requested")
             self.force_enrich()
             return True
+        print("Enrichment not requested as document was enriched recently")
         return False
-
-    @cached_property
-    def can_enrich(self) -> bool:
-        """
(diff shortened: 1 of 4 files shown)


src/caselawclient/models/documents/__init__.py
```diff
@@ -347,18 +347,13 @@ def enrich(self) -> bool:
...
-        if self.can_enrich:
+        if self.enriched_recently is False:
+            print("Enrichment requested")
             self.force_enrich()
             return True
+        print("Enrichment not requested as document was enriched recently")
         return False
 
-    @cached_property
-    def can_enrich(self) -> bool:
```
dragon-dxw, line 356: A breaking change to the API, but appears safe.

dragon-dxw (commented):
I'd like Nick to take a look at our approach.

I anticipate this will significantly increase the number of errors in the alerts channel, if we haven't already changed the downstream error checking (enrichment, privAPI). Alert blindness is already an issue.

No objections in principle but think it's worth sitting on this for a little while until other things are in place.

src/caselawclient/models/documents/__init__.py
anthonyhashemi: ah yeah added line in changelog, but did check it is unused in eui, pui

anthonyhashemi:
> I'd like Nick to take a look at our approach.
> 
> I anticipate this will significantly increase the number of errors in the alerts channel, if we haven't already changed the downstream error checking (enrichment, privAPI). Alert blindness is already an issue.
> 
> No objections in principle but think it's worth sitting on this for a little while until other things are in place.

This should have no effect other than allowing us to try to enrich xml that was previously being blocked due to the xml not being valid against the schema as discussed (detailed in the linked ticket and related epic). I do not plan to bulk enrich to start with. I will run enrichment on individual files in the re-enrich pile from the beginning of Feb when the pipeline went down and see how it goes, and if it looks ok, I will just finish those from this Feb and leave the ones from before, so I do not expect a large amount of errors in the alerts channel.

PR #913, opened 2025-02-27
Provide bytestrings when we're producing search XML.

dragon-dxw:
[Changes to the search](https://github.com/nationalarchives/ds-caselaw-marklogic/pull/158) may mean that search will return XML with an encoding declaration, but the way we absorb that XML in the API client is incorrect -- we get it as a `str` which means that it's already abstract unicode rather than undecoded bytes, and lxml correctly (and annoyingly) tells us it doesn't want to parse abstract unicode as if it's undecoded UTF-8.

The solution is to return bytestrings, which so long as consumers aren't looking too carefully into the types it gets from the API client and just passing them to lxml or similar should be absolutely fine.

This should be safe to run with or without the marklogic changes which depend on them.

This code runs with no errors to demonstrate the problem:

```python
import lxml.etree
import pytest

raw_bytes = b"<x/>"
raw_string = "<x/>"
declared_bytes = b'<?xml version="1.0" encoding="utf-8"?><x/>'
declared_string = '<?xml version="1.0" encoding="utf-8"?><x/>'

lxml.etree.fromstring(raw_bytes)
lxml.etree.fromstring(raw_string)
lxml.etree.fromstring(declared_bytes)
with pytest.raises(Exception):
    lxml.etree.fromstring(declared_string)
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -55,22 +55,22 @@ def generate_search_response_xml_fixture() -> Callable:
         Callable: Function that generates search response XML.
     """
 
-    def _generate_search_response_xml(response_content: str, facets="") -> str:
+    def _generate_search_response_xml(response_content: str, facets="") -> bytes:
         """
-        Generate a search response XML string.
+        Generate a search response XML bytestring.
 
         Args:
             response_content (str): Content of the search response.
 
         Returns:
-            str: Generated search response XML string.
+            bytes: Generated search response XML bytestring.
         """
         return (
             '<search:response xmlns:search="http://marklogic.com/appservices/search" total="2">'
             f"{response_content}"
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -55,7 +55,7 @@ def generate_search_response_xml_fixture() -> Callable:
         Callable: Function that generates search response XML.
     """
 
-    def _generate_search_response_xml(response_content: str, facets="") -> str:
+    def _generate_search_response_xml(response_content: str, facets="") -> bytes:
         """
         Generate a search response XML string.
```
jacksonj04, line 60: s/string/bytestring

jacksonj04 (approved):

PR #916, opened 2025-02-28
Return identifiers in search results

dragon-dxw:
`SearchResults.identifiers` returns a list of identifiers.

Could do with e2e tests at some point in the future once everything leading up to this point is pushed up; confirming all the different components are working together.

Contains a commit from a previous PR because it depends on that work.

### DIFF
--- a/src/caselawclient/responses/search_result.py
+++ b/src/caselawclient/responses/search_result.py
@@ -179,6 +179,13 @@ class SearchResult:
         identifiers_etree = None if not identifiers_etrees else identifiers_etrees[0]
         return unpack_all_identifiers_from_etree(identifiers_etree)
 
+    @cached_property
+    def slug(self) -> str:
+        preferred = self.identifiers.preferred()
+        if not preferred:
+            raise RuntimeError("No preferred identifier for search result")
+        return str(preferred.url_slug)
+
     @property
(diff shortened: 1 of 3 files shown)


jacksonj04 (changes requested):
I think this just has one missing (but important) test case for "a search result has been returned which doesn't have a preferred identifier".

src/caselawclient/responses/search_result.py
```diff
@@ -169,6 +171,21 @@ def uri(self) -> DocumentURIString:
             self._get_xpath_match_string("@uri").lstrip("/").split(".xml")[0],
         )
 
+    @property
+    def identifiers(self) -> Identifiers:
+        identifiers_etrees = self._get_xpath("//identifiers")
+        if count := len(identifiers_etrees) != 1:
+            logging.warning(f"{count} //identifiers nodes found in search result, expected 1.")
+        identifiers_etree = None if not identifiers_etrees else identifiers_etrees[0]
+        return unpack_all_identifiers_from_etree(identifiers_etree)
```
jacksonj04, line 180: I knew there was a reason we made this independent of having a `Document`!

jacksonj04 (approved):
🚢

PR #923, opened 2025-03-05
⚠️ [FCL-735] Inserting a new document requires an explicit document type collection

jacksonj04:
## Summary of changes

This removes the reliance on a document's URI to determine its type collection in MarkLogic. This is now derived directly from the XML itself.

BREAKING CHANGE: `insert_document_xml()` now requires a Document type class be provided

## Checklist

### DIFF
--- a/tests/client/test_save_copy_delete_judgment.py
+++ b/tests/client/test_save_copy_delete_judgment.py
@@ -11,6 +11,7 @@ from caselawclient.Client import ROOT_DIR, MarklogicApiClient
 from caselawclient.client_helpers import VersionAnnotation, VersionType
 from caselawclient.errors import InvalidContentHashError
 from caselawclient.models.documents import DocumentURIString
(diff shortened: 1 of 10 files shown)


tests/client/test_save_copy_delete_judgment.py
```diff
@@ -143,9 +145,10 @@ def test_insert_document_xml(self):
                 ),
             }
             self.client.insert_document_xml(
-                uri,
-                document_xml,
-                VersionAnnotation(
+                document_uri=uri,
+                document_xml=document_xml,
+                document_type=Judgment,
```
dragon-dxw, line 150: This feels a little like it might want to be a Judgment.insert_document_xml in future.

dragon-dxw (approved):

dragon-dxw:
The breaking change should only break the ingester (which we want to fix) (and the apiclient, which is handled here)

PR #930, opened 2025-03-06
Ensure that identifiers are only got for the current search result.

dragon-dxw:
`doc.xpath("//cat").xpath("//dog")` returns all `dog` nodes, not those that are the descendants of a `cat` node.

To specify that, say `.//dog` instead.


tests/responses/test_search_response.py
```diff
@@ -15,6 +15,48 @@ def setup_method(self):
...
+                </search:extracted>
+                </search:result>
+                <search:result xmlns:search="http://marklogic.com/appservices/search" uri="/uksc/2015/20.xml">
+                <search:extracted kind="identifiers">
+                <identifiers>
+                    <identifier>
+                        <namespace>ukncn</namespace>
+                        <uuid>2d80bf1d-e3ea-452f-965c-041f4399f2dd</uuid>
+                        <value>[1901] UKSC 4444</value>
+                        <url_slug>uksc/1901/2222</url_slug>
```
jacksonj04, line 48: Can we make this match the actual expected slug for the value, so it's closer to reality?

jacksonj04 (approved):
One small change otherwise send it.

PR #984, opened 2025-04-17
[FCL-799] Reinstate PRIVATE_ASSET_BUCKET_REGION

dragon-dxw:
## Summary of changes

Reinstate environment variable so AWS knows what region it's in
Also bumps version

https://national-archives.atlassian.net/browse/FCL-799

## Checklist

### DIFF
--- a/src/caselawclient/models/utilities/aws.py
+++ b/src/caselawclient/models/utilities/aws.py
@@ -50,6 +50,7 @@ def create_aws_client(service: Literal["s3", "sns"]) -> Any:
     aws = boto3.session.Session()
     return aws.client(
         service,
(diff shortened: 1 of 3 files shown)


src/caselawclient/models/utilities/aws.py
```diff
@@ -50,6 +50,7 @@ def create_aws_client(service: Literal["s3", "sns"]) -> Any:
     aws = boto3.session.Session()
     return aws.client(
         service,
+        region_name=env("PRIVATE_ASSET_BUCKET_REGION", default=None),
```
rjw1, line 53: I think you could safely default to `eu-west-2`

PR #996, opened 2025-04-23
[FCL-818] Make sure values are validated when Identifier objects are created

jacksonj04:
## Summary of changes

We aren't currently checking that identifier values are valid when we try to initialise the object. We should do so, so that we can't then save invalid identifier values to the database.

This PR adds running the `schema.validate_identifier()` method to the `__init__` method for `Identifier`s (with attendant test changes, now that test identifiers must also be valid).

## Jira

FCL-818

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,19 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog 1.0.0].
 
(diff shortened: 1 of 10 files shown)


dragon-dxw (approved):

dragon-dxw:
took me a while to parse tn4t35t5

PR #1010, opened 2025-04-30
[FCL-833] Improve Search factory output

jacksonj04:
Make the `SearchResultFactory` defaults more closely match the output of a real search, so we can use them downstream in other services as a reliable stand-in.

## Jira

FCL-833

### DIFF
--- a/src/caselawclient/factories.py
+++ b/src/caselawclient/factories.py
@@ -146,15 +146,18 @@ class SearchResultMetadataFactory(SimpleFactory):
 class SearchResultFactory(SimpleFactory):
     target_class = SearchResult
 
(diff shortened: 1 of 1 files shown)


src/caselawclient/factories.py
```diff
@@ -148,13 +148,17 @@ class SearchResultFactory(SimpleFactory):
 
     # "name_of_attribute": ("name of incoming param", "default value")
     PARAMS_MAP = {
-        "uri": "test/2023/123",
+        "uri": "d-6323a233-4e96-4b60-b49c-6ca0e78fac3a",
```
dragon-dxw, line 151: I like having these distinctively obviously test data, but this is fine.

dragon-dxw (approved):

PR #1037, opened 2025-05-14
[FCL-49] Add supporting functions for issuing FCLIDs to all documents

jacksonj04:
## Summary of changes

To support being able to issue FCLIDs to all documents without them:

- Make safe FCLID minting its own function
- Add query to retrieve documents without an FCLID

## Checklist

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -1198,6 +1198,24 @@ class MarklogicApiClient:
 
         return results
 
+    def get_missing_fclid(
+        self,
+        maximum_records: int = 50,
+    ) -> list[str]:
+        """Retrieve documents which do not have an FCLID."""
+        vars: query_dicts.GetMissingFclidDict = {
+            "maximum_records": maximum_records,
+        }
+
+        results: list[str] = get_multipart_strings_from_marklogic_response(
+            self._send_to_eval(
+                vars,
+                "get_missing_fclid.xqy",
+            )
+        )
+
+        return results
+
     def resolve_from_identifier_slug(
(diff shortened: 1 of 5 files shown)


src/caselawclient/models/documents/__init__.py
```diff
@@ -378,6 +378,13 @@ def validates_against_schema(self) -> bool:
         """
         return self.api_client.validate_document(self.uri)
 
+    def mint_fclid(self) -> None:
+        """Mint a new FCLID if needed, and save."""
+        if len(self.identifiers.of_type(FindCaseLawIdentifier)) < 1:
```
dragon-dxw, line 383: `== 0` is probably clearer?

src/caselawclient/models/documents/__init__.py
```diff
@@ -378,6 +378,13 @@ def validates_against_schema(self) -> bool:
         """
         return self.api_client.validate_document(self.uri)
 
+    def mint_fclid(self) -> None:
```
dragon-dxw, line 381: Not sure this is a great name. "ensure_has_fclid"? (not sure I love that either)

src/caselawclient/Client.py
```diff
@@ -1198,6 +1198,24 @@ def get_recently_parsed(
 
         return results
 
+    def get_missing_fclid(
+        self,
+        maximum_records: int = 50,
+    ) -> list[str]:
+        """Retrieve documents which do not have an FCLID."""
```
dragon-dxw, line 1205: Retrieve **published** documents.

dragon-dxw (approved):

PR #1051, opened 2025-05-20
Added has_html_representation to Document

anthonyhashemi:
https://national-archives.atlassian.net/browse/FCL-558

## Summary of changes

Added has_html_representation to Document

have tested the xquery locally

## Checklist

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -889,6 +889,13 @@ class MarklogicApiClient:
             "%Y-%m-%dT%H:%M:%S.%f%z",
         )
 
+    def has_html_representation(self, judgment_uri: DocumentURIString) -> bool:
+        uri = self._format_uri_for_marklogic(judgment_uri)
+        vars: query_dicts.HasHtmlRepresentationDict = {
+            "uri": uri,
+        }
+        return self._eval_and_decode(vars, "has_html_representation.xqy") == "true"
+
     def set_property(
         self,
         judgment_uri: DocumentURIString,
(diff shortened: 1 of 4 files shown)


src/caselawclient/Client.py
```diff
@@ -894,6 +894,13 @@ def get_version_created_datetime(self, judgment_uri: DocumentURIString) -> datet
             "%Y-%m-%dT%H:%M:%S.%f%z",
         )
 
+    def has_html_representation(self, judgment_uri: DocumentURIString) -> bool:
```
anthonyhashemi, line 897: want to write tests for the xqy / an integration test rather than just a test that mocks the call to marklogic but I will do that in a specific ticket as we dont have that setup yet. Will add that mocking test though

jacksonj04: Yeah, we still haven't built a good framework for testing our MarkLogic queries are doing what we actually expect.

src/caselawclient/Client.py
```diff
@@ -889,6 +889,13 @@ def get_version_created_datetime(self, judgment_uri: DocumentURIString) -> datet
...
+        uri = self._format_uri_for_marklogic(judgment_uri)
+        vars: query_dicts.HasHtmlRepresentationDict = {
+            "uri": uri,
+        }
+        return self._eval_and_decode(vars, "has_html_representation.xqy") == "true"
```
jacksonj04, line 897: I don't think this file has been checked in?

jacksonj04 (changes requested):
This is missing the actual query to be run!

src/caselawclient/Client.py
anthonyhashemi: it got lost in a sea of other unstaged files I have 🤦

PR #1058, opened 2025-05-22
Update toml to version 37.3.0

anthonyhashemi:
## Summary of changes

Update toml to version 37.3.0


jacksonj04 (changes requested):
This needs the changelog updating as well.

PR #1066, opened 2025-05-28
[FCL-881] Improve behaviour of document versions

jacksonj04:
## Checklist


src/caselawclient/models/utilities/__init__.py
```diff
@@ -3,23 +3,22 @@
 
 from requests_toolbelt.multipart.decoder import BodyPart
 
-VERSION_REGEX = r"xml_versions/(\d{1,10})-(\d{1,10}|TDR)"
+from caselawclient.types import DocumentURIString, MarkLogicDocumentURIString
+
+VERSION_REGEX = r"xml_versions/(\d{1,10})-"
```
dragon-dxw, line 8: Since we've dropped the document reference bit of the regex, we probably want to drop that from the explanation below.

dragon-dxw (approved):

PR #1082, opened 2025-06-05
[FCL-885] Identifier types are able to express constraints and those constraints are enforced at the API Client layer

jacksonj04:
## Summary of changes

This PR provides for identifiers (or, more specifically, identifier _schemas_) being able to express constraints on where they can be used.

### New constraints

- Provides a new flag for indicating to downstream applications that manual editing of a schema shouldn't be permitted
- Provides a new flag and method for validating that an identifier is globally unique
- Identifier schemas can now specify which types of document they are valid for

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,18 +8,33 @@ The format is based on [Keep a Changelog 1.0.0].
 
 ### BREAKING CHANGE
 
+- `Identifiers.validate()` is now `Identifiers.validate_uuids_match_keys()`
 - `IdentifierSchema.validate_identifier` has been renamed `IdentifierSchema.validate_identifier_value`
 
 ### Feat
 
+- **Document**: saving identifiers now correctly raises exception when validation fails
+- move to using structured objects to represent validation failures rather than exceptions
+- **Identifiers**: validating identifiers now performs a full check of all individual identifiers
+- **Identifier**: add ability to validate that an identifier is acceptable for a given document class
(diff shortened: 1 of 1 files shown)


tests/models/documents/test_document_verbs.py
```diff
@@ -14,27 +14,55 @@
...
+from caselawclient.types import SuccessFailureMessageTuple
 
 
 class TestDocumentSaveIdentifiers:
     def test_document_save_identifiers(self, mock_api_client):
         """
         given a particular Document with a known value of Identifiers (probably mock this out tbh)
         when I call document.save_identifiers() it
-        calls identifiers.validate and
+        calls identifiers.perform_all_validations and
```
dragon-dxw, line 29: "validates it and"

tests/models/identifiers/test_identifier_constraints.py
```diff
@@ -0,0 +1,203 @@
...
+
+class TestLimitedToJudgmentDocumentTypeIdentifier(TestIdentifier):
+    schema = TestLimitedToJudgmentDocumentTypeIdentifierSchema
+
+
+class TestCorrectDocumentType(Document):
+    __test__ = False
+
+
+class TestIncorrectDocumentType(Document):
```
dragon-dxw, line 38: Feels odd that this is the same as the Corect one?

dragon-dxw (approved):

PR #1097, opened 2025-06-16
[FCL-889] Don't enrich if no content

dragon-dxw:
## Summary of changes

If a document doesn't have a docx, it should not be enriched; when enrichment is attempted, fail gracefully but record that an attempt was made to ensure it doesn't block other enrichment.

Difficult to manually test locally -- worth additional scrutiny in staging?

https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-889

## Checklist

### DIFF
--- a/src/caselawclient/models/documents/__init__.py
+++ b/src/caselawclient/models/documents/__init__.py
@@ -34,7 +34,7 @@ from caselawclient.models.utilities.aws import (
 from caselawclient.types import DocumentURIString
 
 from .body import DocumentBody
-from .exceptions import CannotPublishUnpublishableDocument, DocumentNotSafeForDeletion
+from .exceptions import CannotEnrichUnenrichableDocument, CannotPublishUnpublishableDocument, DocumentNotSafeForDeletion
 from .statuses import DOCUMENT_STATUS_HOLD, DOCUMENT_STATUS_IN_PROGRESS, DOCUMENT_STATUS_NEW, DOCUMENT_STATUS_PUBLISHED
 
 MINIMUM_ENRICHMENT_TIME = datetime.timedelta(minutes=20)
@@ -341,22 +341,34 @@ class Document:
             now.isoformat(),
         )
 
+        if not self.can_enrich:
+            msg = f"{self.uri} has no docx"
+            raise CannotEnrichUnenrichableDocument(msg)
+
         announce_document_event(
             uri=self.uri,
             status="enrich",
             enrich=True,
         )
 
-    def enrich(self) -> bool:
+    def enrich(self, even_if_recent: bool = False, accept_failures: bool = False) -> bool:
         """
         Request enrichment of a document, if it's sensible to do so.
         """
-        if self.enriched_recently is False:
-            print("Enrichment requested")
+        if not (even_if_recent) and self.enriched_recently:
+            print("Enrichment not requested as document was enriched recently")
+            return False
+
+        print("Enrichment requested")
+
+        try:
             self.force_enrich()
-            return True
-        print("Enrichment not requested as document was enriched recently")
-        return False
+        except CannotEnrichUnenrichableDocument as e:
+            if not accept_failures:
+                raise e
+            return False
+
+        return True
 
     @cached_property
     def enriched_recently(self) -> bool:
@@ -499,6 +511,13 @@ class Document:
     def can_reparse(self) -> bool:
         """
         Is it sensible to reparse this document?
+        """
+        return self.docx_exists()
+
+    @cached_property
+    def can_enrich(self) -> bool:
+        """
+        Is it possible to enrich this document?
         """
         return self.docx_exists()
 
--- a/src/caselawclient/models/documents/__init__.py
+++ b/src/caselawclient/models/documents/__init__.py
@@ -342,7 +342,7 @@ class Document:
         )
 
         if not self.can_enrich:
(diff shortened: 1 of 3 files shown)


src/caselawclient/models/documents/__init__.py
```diff
@@ -329,31 +329,39 @@ def status(self) -> str:
 
         return DOCUMENT_STATUS_NEW
 
-    def force_enrich(self) -> None:
+    def force_enrich(self) -> bool:
```
dragon-dxw, line 332: This is technically an API change, but nothing should care about something that doesn't return a value returning something.

src/caselawclient/models/documents/__init__.py
```diff
@@ -501,6 +509,13 @@ def can_reparse(self) -> bool:
         """
         return self.docx_exists()
 
+    @cached_property
+    def can_enrich(self) -> bool:
```
anthonyhashemi, line 513: `is_sensible_to_enrich` could be clearer than `can_enrich` but not a big deal

src/caselawclient/models/documents/__init__.py
anthonyhashemi: yeah, technically it is a breaking change, so just remember to reflect in the version bump

src/caselawclient/models/documents/__init__.py
```diff
@@ -329,31 +329,39 @@ def status(self) -> str:
...
+        # We save the time before checking for docx so if there isn't a
+        # docx we will skip this document
         self.api_client.set_property(
             self.uri,
             "last_sent_to_enrichment",
             now.isoformat(),
         )
 
+        if not (self.can_enrich):
+            print("Enrichment of {self} rejected")
```
jacksonj04, line 347: This feels like it should be an exception. Something is trying to force an enrichment when it shouldn't, and we have a method in `can_enrich` to check if it should be possible beforehand.

jacksonj04 (changes requested):

src/caselawclient/models/documents/__init__.py
```diff
@@ -340,22 +340,34 @@ def force_enrich(self) -> None:
             now.isoformat(),
         )
 
+        if not self.can_enrich:
+            msg = f"{self.uri} has no docx"
```
jacksonj04, line 344: This message is incorrect:

```suggestion
            msg = f"{self.uri} has no body content"
```

src/caselawclient/models/documents/exceptions.py
```diff
@@ -2,5 +2,9 @@ class CannotPublishUnpublishableDocument(Exception):
     """A document which has failed publication safety checks in `Document.is_publishable` cannot be published."""
 
 
+class CannotEnrichUnenrichableDocument(Exception):
+    """We tried to enrich a document without a docx"""
```
jacksonj04, line 6: I don't think we should be describing the logic of the check in here; this should be something like:

```suggestion
    """A document which is not in an enrichable state as per `Document.can_enrich` cannot be enriched"""
```

tests/models/documents/test_document_verbs.py
```diff
@@ -142,6 +144,60 @@ def test_force_enrich(self, mock_announce_document_event, mock_api_client):
             enrich=True,
         )
 
+    @time_machine.travel(datetime.datetime(1955, 11, 5, 6))
+    @patch("caselawclient.models.documents.announce_document_event")
+    @patch("caselawclient.models.documents.Document.can_enrich", new_callable=PropertyMock, return_value=False)
+    def test_force_enrich_but_no_docx(self, mock_can_enrich, mock_announce_document_event, mock_api_client):
```
jacksonj04, line 150: Referring to docx again

tests/models/documents/test_document_verbs.py
```diff
@@ -142,6 +144,60 @@ def test_force_enrich(self, mock_announce_document_event, mock_api_client):
...
+        )
+
+        mock_announce_document_event.assert_not_called()
+
+
+class TestDocumentEnrich:
+    @time_machine.travel(datetime.datetime(1955, 11, 5, 6))
+    @patch("caselawclient.models.documents.announce_document_event")
+    @patch("caselawclient.models.documents.Document.can_enrich", new_callable=PropertyMock, return_value=False)
+    def test_enrich_with_no_docx(self, mock_can_enrich, mock_announce_document_event, mock_api_client):
```
jacksonj04, line 169: More docx references...

tests/models/documents/test_document_verbs.py
```diff
@@ -142,6 +144,60 @@ def test_force_enrich(self, mock_announce_document_event, mock_api_client):
...
+            "last_sent_to_enrichment",
+            "1955-11-05T06:00:00+00:00",
+        )
+
+        mock_announce_document_event.assert_not_called()
+
+    @time_machine.travel(datetime.datetime(1955, 11, 5, 6))
+    @patch("caselawclient.models.documents.announce_document_event")
+    @patch("caselawclient.models.documents.Document.can_enrich", new_callable=PropertyMock, return_value=False)
+    def test_enrich_with_no_docx_and_exception_ignored(
```
jacksonj04, line 186: More docx...

jacksonj04 (approved):
Just some tiny naming (docx vs content) changes to make, otherwise LGTM

PR #1113, opened 2025-07-02
[FCL-880] Support direct manipulation of identifiers list

jacksonj04:
## Summary of changes

Changes to the API client to support the EUI being able to directly manipulate the list of identifiers.

## Checklist

## Jira

- FCL-880
- FCL-994

### DIFF
--- a/src/caselawclient/models/identifiers/collection.py
+++ b/src/caselawclient/models/identifiers/collection.py
@@ -5,7 +5,6 @@ from lxml import etree
 from caselawclient.types import SuccessFailureMessageTuple
 
 from . import Identifier
-from .exceptions import UUIDMismatchError
 from .fclid import FindCaseLawIdentifier
 from .neutral_citation import NeutralCitationNumber
 from .press_summary_ncn import PressSummaryRelatedNCNIdentifier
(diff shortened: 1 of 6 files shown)


src/caselawclient/models/identifiers/collection.py
```diff
@@ -22,20 +21,54 @@
 
 
 class IdentifiersCollection(dict[str, Identifier]):
-    def validate_uuids_match_keys(self) -> None:
+    def validate_uuids_match_keys(self) -> SuccessFailureMessageTuple:
         for uuid, identifier in self.items():
             if uuid != identifier.uuid:
-                msg = "Key of {identifier} in Identifiers is {uuid} not {identifier.uuid}"
-                raise UUIDMismatchError(msg)
+                return SuccessFailureMessageTuple(
```
dragon-dxw, line 27: It's possible that having SuccessMessageTuple and FailureMessageTuple child classes would improve readability.

src/caselawclient/models/identifiers/collection.py
```diff
@@ -22,20 +21,54 @@
...
+                    False, [f"Key of {identifier} in Identifiers is {uuid} not {identifier.uuid}"]
+                )
+
+        return SuccessFailureMessageTuple(True, [])
+
+    def check_only_single_current_identifier_per_type(self) -> SuccessFailureMessageTuple:
```
dragon-dxw, line 33: This will probably want tweaking when we might have multiple identifiers per document; is it worth changing the name to reflect this to avoid API changes later?

`check_no_multiple_identifiers_for_solo_identifier_types`?

dragon-dxw (approved):

PR #1114, opened 2025-07-02
[FCL-1007] Refactor identifier collections

jacksonj04:
## Summary of changes

Move where we define the concept of a collection of identifiers around, so we can avoid circular import problems in future.

## Checklist


anthonyhashemi (approved):
Does what it says on the tin 👌

PR #1126, opened 2025-07-08
Use `/id/doc` namespace for work with no `tna.` prefix

dragon-dxw:
## Summary of changes

We agreed now is a good time to migrate to `/id/doc/tn4t35ts` as a URI.
https://dxw.slack.com/archives/C04FQ3GFGUQ/p1751972123605659

## Checklist

### DIFF
--- a/src/caselawclient/models/documents/__init__.py
+++ b/src/caselawclient/models/documents/__init__.py
@@ -570,7 +570,7 @@ class Document:
     def xml_with_correct_frbr(self) -> bytes:
         """Dynamically modify FRBR uris to reflect current storage location and FCL id"""
         fcl_identifiers = self.identifiers.of_type(FindCaseLawIdentifier)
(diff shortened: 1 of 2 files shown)


src/caselawclient/models/documents/__init__.py
```diff
@@ -570,7 +570,7 @@ def content_as_html(self) -> str | None:
     def xml_with_correct_frbr(self) -> bytes:
         """Dynamically modify FRBR uris to reflect current storage location and FCL id"""
         fcl_identifiers = self.identifiers.of_type(FindCaseLawIdentifier)
-        work_uri = f"https://caselaw.nationalarchives.gov.uk/id/{fcl_identifiers[0].url_slug}"
+        work_uri = f"https://caselaw.nationalarchives.gov.uk/id/doc/{fcl_identifiers[0].value}"
```
jacksonj04, line 573: Seeing this pinned down makes me so happy!

PR #1132, opened 2025-07-10
Don't insert akn namespaces unnecessarily

dragon-dxw:
## Summary of changes

In order to use v1 canonicalisation to support non-messy ref tags, ensure that we don't insert spurious akn namespaces into the middle of the document.

## Checklist


src/caselawclient/xslt/modify_xml_live.xsl
```diff
@@ -16,51 +19,51 @@
     <!-- <xsl:template match="akn:identification/FRBRWork/FRBRthistext/text()"><xsl:copy-of select="$cat" /></xsl:template> -->
 
     <xsl:template match="akn:identification/akn:FRBRWork/akn:FRBRthis">
```
dragon-dxw, line 21: The weirdness of this having to have `akn:` attributes:

https://my.slack.com/archives/C04FQ3GFGUQ/p1752504654822609?thread_ts=1752159098.748679&cid=C04FQ3GFGUQ

> In XSLT 1, which is miserable to write these days, the default namespace you declare on the <xsl:stylesheet> element is not consulted when a template‐match pattern or select expression contains an un-prefixed name test. In XSLT 2, there's a default-xpath-namespace attribute for that

PR #1139, opened 2025-07-14
Bandit

dragon-dxw:
## Summary of changes

* use defusedxml instead of xml (perhaps we should not be using the `xml` library at all)
* ignore a requests timeout in tests about a timeout
* ignore asserts in smoketests

## Checklist


dragon-dxw:
https://dxw.slack.com/archives/C04FQ3GFGUQ/p1752504806313179?thread_ts=1752500939.154289&cid=C04FQ3GFGUQ

PR #1144, opened 2025-07-18
Upgrade to lxml 6

dragon-dxw:
## Summary of changes

Changelog: https://github.com/lxml/lxml/blob/master/CHANGES.txt

Nothing looks controversial; the only real breaking change appears to be removal of Python 3.7 support

Should probably wait for editor to be upgraded first, but good to run tests on API client etc.

## Checklist


anthonyhashemi (approved):
sweet

PR #1151, opened 2025-07-28
Fix document.force_reparse for parser failures

anthonyhashemi:
## Summary of changes

Fix document.force_reparse for parser failures

## Checklist


anthonyhashemi:
There's an issue with uploading codecoverage to codeclimate for python 3.13 but all tests pass and works on the rest of them. We are moving away from codeclimate anyway.

PR #1155, opened 2025-07-30
FCL-1063: Add ability to compare documents

dragon-dxw:
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?jql=assignee%20%3D%2060bdecb0c90cb2006806ae0f&selectedIssue=FCL-1063

## Summary of changes

`comparison = doc.compare_to(other_doc)`

`comparison.match()` is false

```
{
  'court': {'label': 'court', 'this_value': 'Court of Testing', 'that_value': 'Different', 'match': False}, 
  'date': {'label': 'date', 'this_value': '2023-02-03', 'that_value': '2000-01-09', 'match': False}, 
  'name': {'label': 'name', 'this_value': 'Judgment v Judgement', 'that_value': 'Different', 'match': False}
}
```

## Checklist

### DIFF
--- a/src/caselawclient/models/documents/comparison.py
+++ b/src/caselawclient/models/documents/comparison.py
@@ -6,50 +6,37 @@ import caselawclient.models.documents
 class AttributeComparison(TypedDict):
     """Results from a comparison of one attribute across two documents"""
 
-    attribute: str
-    attribute_label: str
+    label: str
     this_value: str
     that_value: str
     match: bool
 
 
-class Comparison(list[AttributeComparison]):
+class Comparison(dict[str, AttributeComparison]):
     def __init__(
         self, this_doc: "caselawclient.models.documents.Document", that_doc: "caselawclient.models.documents.Document"
     ):
         # court, date, title
-        self.append(
-            AttributeComparison(
-                attribute="court",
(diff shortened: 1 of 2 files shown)


src/caselawclient/models/documents/__init__.py
```diff
@@ -583,3 +584,6 @@ def xml_with_correct_frbr(self) -> bytes:
         return self.body.apply_xslt(
             "modify_xml_live.xsl", work_uri=work_uri, expression_uri=expression_uri, manifestation_uri=manifestation_uri
         )
+
+    def compare_to(self, that_doc: "Document") -> comparison.Comparison:
```
anthonyhashemi, line 588: To keep things simple what do you think about:

```
   def is_likely_a_version_of(self, document):
        """
        Returns True if this document matches the other document
        on comparison fields: court, name, date. This method is not a foolproof way of deciding if a document is a version of another but it uses these comparison fields as a heuristic.
        """
        comparison_fields = ["court", "name", "date"]
        return all(getattr(self, f) == getattr(document, f) for f in comparison_fields)
```

anthonyhashemi (approved):

PR #1159, opened 2025-07-31
Make the package version number in the docs dynamic

jacksonj04:
## Summary of changes

Previously, the generated documentation was using a static version number for the library (v12). Make it dynamic, so people blindly copying and pasting will always get the latest version.

## Checklist

### DIFF
--- a/src/caselawclient/__init__.py
+++ b/src/caselawclient/__init__.py
@@ -1,4 +1,21 @@
-"""
+import os
+
+if os.getenv("PDOC_DYNAMIC_VERSION") == "1":
+    from pathlib import Path
+
+    import tomllib
+
+    pyproject_path = Path(__file__).parent.parent.parent / "pyproject.toml"
+    with pyproject_path.open("rb") as f:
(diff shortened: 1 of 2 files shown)


src/caselawclient/__init__.py
```diff
@@ -15,7 +29,7 @@
 or in your projects `requirements.txt` with:
 
 ```text
-ds-caselaw-marklogic-api-client~=12.0.0
+ds-caselaw-marklogic-api-client{__version_string__}
```
dragon-dxw, line 32: feels weird that the version string has the `~=` prefix, but not a problem

dragon-dxw (approved):

src/caselawclient/__init__.py
dragon-dxw: A quicker fix would have been just removing the version number entirely?
And we should probably move to poetry long term.

jacksonj04: We should, but this is included for those who haven't. I've fixed up how we define this.

PR #1167, opened 2025-08-04
Add timeout for marklogic eval

dragon-dxw:
https://national-archives.atlassian.net/jira/software/projects/FCL/boards/136?selectedIssue=FCL-1056
## Summary of changes

Adds optional environment variables `CONNECT_TIMEOUT` and `READ_TIMEOUT` with default values borrowed from https://requests.readthedocs.io/en/latest/user/advanced/#timeouts (slightly more than a TCP retry window and slightly less than a default gunicorn timeout, respectively)

## Checklist

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -59,7 +59,7 @@ env = environ.Env()
 
 # Requests timeouts: https://requests.readthedocs.io/en/latest/user/advanced/
 CONNECT_TIMEOUT = float(os.environ.get("CONNECT_TIMEOUT", "3.05"))
-READ_TIMEOUT = float(os.environ.get("READ_TIMEOUT", "27"))
+READ_TIMEOUT = float(os.environ.get("READ_TIMEOUT", "10.0"))
 
 ROOT_DIR = os.path.dirname(os.path.realpath(__file__))
 DEFAULT_XSL_TRANSFORM = "accessible-html.xsl"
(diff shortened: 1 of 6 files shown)


src/caselawclient/Client.py
```diff
@@ -743,6 +748,7 @@ def eval(
             url=self._path_to_request_url(path),
             headers=headers,
             data=data,
+            timeout=(CONNECT_TIMEOUT, READ_TIMEOUT),
```
jacksonj04, line 751: Do we use eval for write operations which may potentially take longer? Is it worth making these parameters which just default to the constants?

src/caselawclient/Client.py
```diff
@@ -56,6 +56,11 @@
 )
 
 env = environ.Env()
+
+# Requests timeouts: https://requests.readthedocs.io/en/latest/user/advanced/
+CONNECT_TIMEOUT = float(os.environ.get("CONNECT_TIMEOUT", "3.05"))
+READ_TIMEOUT = float(os.environ.get("READ_TIMEOUT", "27"))
```
jacksonj04, line 62: This can be a lot more aggressive I think. If we haven't had a response back in... maybe five seconds we should consider showing the user an error.

If we catch timeout errors and re-raise as something more specific to us timing out when talking to MarkLogic then we can get a better picture in Rollbar as well.

anthonyhashemi (approved):
LGTM

PR #1194, opened 2025-08-15
FCL 1132: Block enrichment and reparsing if tribunal backlog

dragon-dxw:
## Summary of changes

https://national-archives.atlassian.net/browse/FCL-1132

Using the presence of `uk:party` as a temporary measure before building a proper way of dealing with the additional data, block enrichment and reparsing of documents from the tribunal backlog.

Reparsing is definately needed, enrichment might be safe (but pointless)

### DIFF
--- a/src/caselawclient/models/documents/body.py
+++ b/src/caselawclient/models/documents/body.py
@@ -186,6 +186,13 @@ class DocumentBody:
 
         return False
 
+    @cached_property
+    def has_external_data(self) -> bool:
+        """Is there data which is not present within the source document:
+        is there a spreadsheet which has populated some fields. The current implementation
+        "is there a uk:party tag" is intended as a stopgap whilst we're not importing that data."""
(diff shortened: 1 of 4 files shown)


src/caselawclient/models/documents/body.py
```diff
@@ -144,6 +144,13 @@ def has_content(self) -> bool:
 
         return False
 
+    @cached_property
+    def has_external_data(self) -> bool:
+        """Is there data which is not present within the source document:
+        is there a spreadsheet which has populated some fields. The current implementation
+        "is there a uk:party tag" is intended as a stopgap whilst we're not importing that data."""
+        return bool(self._xml.xml_as_tree.xpath("//uk:party", namespaces=DEFAULT_NAMESPACES))
```
anthonyhashemi, line 152: is this the best proxy we have? If so we will have to be very careful anytime change is made to the parser to make sure this is still the case. I want to spend a little bit more time thinking if we have an alternative

PR #1209, opened 2025-08-21
FCL-890 | add extra categories to document body

jlhdxw:
## Summary of changes

Adds the subcategory, secondary category and secondary subcategory to the document body.

## Checklist

### DIFF
--- a/src/caselawclient/models/documents/body.py
+++ b/src/caselawclient/models/documents/body.py
@@ -53,7 +53,27 @@ class DocumentBody:
 
     @cached_property
     def category(self) -> Optional[str]:
-        return self.get_xpath_match_string("/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:category/text()")
+        return self.get_xpath_match_string(
+            "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:category[not(@parent)][1]/text()"
+        )
+
+    @cached_property
+    def secondary_category(self) -> Optional[str]:
+        return self.get_xpath_match_string(
+            "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:category[not(@parent)][2]/text()"
+        )
+
+    @cached_property
+    def subcategory(self) -> Optional[str]:
+        return self.get_xpath_match_string(
+            "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:category[@parent][1]/text()"
+        )
+
+    @cached_property
+    def secondary_subcategory(self) -> Optional[str]:
+        return self.get_xpath_match_string(
+            "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:category[@parent][2]/text()"
+        )
 
     @cached_property
     def case_number(self) -> Optional[str]:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog 1.0.0].
 
 ## Unreleased
 
+### Feat
+
+- **DocumentBody**: update category to match the first category without a parent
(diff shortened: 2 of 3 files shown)


jacksonj04 (changes requested):
We could definitely do with better documenting what on earth these mean!

There's a decent-sized part of me wondering if the right thing to do here is to return an arbitrary data structure under a general `categories` property. Maybe something like

``` python

categories = [
  {
    "name": "Category",
    "subcategories": [
      {
        "name": "Subcategory"
      }
    ]
  },
  {
    "name": "Secondary Category",
    "subcategories": [
      {
        "name": "Secondary Subcategory"
      }
    ]
  }
]

```

which theoretically can scale as far as needed either horizontally or vertically when courts inevitably in future decide that something can have multiple subcategories or similar.

HLLindon:
Just as an aside, the terminology used in teh database doesn't necessarily reflect A) what the databases say on the front end and B) what the courts use or call these data, so they should be taken with a pinch of salt.
For example, one says 'hearingdate' when it is clearly the handed down date - naming things is hard.

src/caselawclient/models/documents/body.py
```diff
@@ -51,9 +56,45 @@ def court(self) -> str:
     def jurisdiction(self) -> str:
         return self.get_xpath_match_string("/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:jurisdiction/text()")
 
+    @cached_property
+    def categories(self) -> List[DocumentCategory]:
```
dragon-dxw, line 60: I think you can just write this as `list[DocumentCategory]` and save yourself the import -- surprised the typechecking didn't change this for you.

src/caselawclient/models/documents/body.py
```diff
@@ -51,9 +56,45 @@ def court(self) -> str:
...
+
+        categories: dict[str, DocumentCategory] = {}
+        children_map: dict[str, list[DocumentCategory]] = {}
+
+        for node in nodes:
+            name = node.text
+            if name is None or not name.strip():
+                continue
+
+            parent = node.get("parent")
```
dragon-dxw, line 72: I'd move this down a couple of lines to put it closer to its use (L77)

dragon-dxw (approved):

src/caselawclient/models/documents/body.py
```diff
@@ -51,9 +56,45 @@ def court(self) -> str:
...
+            if node.get("parent") is None
+            if (name := node.text) and name in categories
+        ]
+
+        return top_level_categories
+
     @cached_property
     def category(self) -> Optional[str]:
-        return self.get_xpath_match_string("/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:category/text()")
+        return self.get_xpath_match_string(
```
jacksonj04, line 95: Does this make conceptual sense any more? I'm slightly erring to deleting it entirely, so downstream things _have_ to handle the possibility of multiple categories, and do it properly.

src/caselawclient/types.py
```diff
@@ -1,4 +1,11 @@
-from typing import NamedTuple
+from dataclasses import dataclass, field
+from typing import List, NamedTuple
```
jacksonj04, line 2: I don't think you need to import the `List` here, you can just use the native `list`.

src/caselawclient/types.py
```diff
@@ -1,4 +1,11 @@
...
+
+
+@dataclass
+class DocumentCategory:
+    name: str
+    subcategories: List["DocumentCategory"] = field(default_factory=list)
```
jacksonj04, line 8: I'm happy with the potential for arbitrary depth here, and that UI will need to either handle it or decide where to cut it off.

jacksonj04 (changes requested):

PR #1230, opened 2025-09-01
Dont allow publishing if content hash is not unique

anthonyhashemi:
## Summary of changes

- Added `api_client.has_unique_content_hash`
- Added `Document.has_unique_content_hash`
- Extended `Document`' `sattributes_to_validate` to consider `has_unique_content_hash`

- All in resulting in `is_publishable` to return False if a document has a content_hash which at least one other document in the database shares.

**Reason:**

So that we block publishing docs which share content hash with another doc, as that implies there is an issue where we need to either delete an entry or merge into another.

## Checklist


jacksonj04 (approved):
LGTM. I wonder if the error message would be better as "there is another document with identical content" or similar, so it's explaining the problem rather than how we detect it, but I don't think it's blocking.

PR #1268, opened 2025-09-18
bump: version 41.1.2 -> 41.1.3

jlhdxw:
### Feat

- **Search**: adds neutral_citation and consignment_number as search options


dragon-dxw (approved):

dragon-dxw:
Possibly should be 41.2?

PR #1294, opened 2025-10-06
[FCL-1325] Add new methods to check if a document is safe to use as a merge source

jacksonj04:
## Summary of changes

The API Client should be able to identify if a source document is a suitable candidate for merging onto a target document. Namely:

- [x] The document only has one version
- [x] The document has never been published
- [ ] The document is in a state where it’s safe to delete

This does not cover working out if a source document is safe to merge given a specific target, only that the source document is in a state where it’s safe to even consider as a merge source.

## Checklist

### DIFF
--- a/src/caselawclient/models/documents/__init__.py
+++ b/src/caselawclient/models/documents/__init__.py
@@ -330,6 +330,16 @@ class Document:
 
         return SuccessFailureMessageTuple(True, [])
 
+    def check_is_safe_to_delete(self) -> SuccessFailureMessageTuple:
+        """Make sure the document is safe to delete."""
+        if not self.safe_to_delete:
+            return SuccessFailureMessageTuple(
+                False,
+                ["This document cannot be deleted because it is published"],
+            )
+
+        return SuccessFailureMessageTuple(True, [])
+
     def check_is_safe_as_merge_source(self) -> SuccessFailureMessageTuple:
         """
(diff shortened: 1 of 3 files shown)


src/caselawclient/models/documents/__init__.py
```diff
@@ -310,6 +310,58 @@ def is_publishable(self) -> bool:
         # An empty list (which is falsy) therefore means the judgment can be published safely.
         return not self.validation_failure_messages
 
+    def check_has_only_one_version(self) -> SuccessFailureMessageTuple:
+        """Make sure the document has exactly one version."""
+        if len(self.versions) > 1:
+            return SuccessFailureMessageTuple(
```
dragon-dxw, line 316: It's probably way too late but I keep thinking that this thing should quack like a boolean plus have additional info if false.

src/caselawclient/models/documents/__init__.py
```diff
@@ -310,6 +310,58 @@ def is_publishable(self) -> bool:
...
+        validations = [
+            self.check_has_only_one_version(),
+            self.check_has_never_been_published(),
+            self.check_is_safe_to_delete(),
+        ]
+
+        success = True
+        messages: list[str] = []
+
+        for validation in validations:
```
dragon-dxw, line 357: This function feels like a thing that should be a method on SuccessFailureMessageTuple.

dragon-dxw (approved):

PR #1509, opened 2026-01-30
FCL-1252: Clean existing S3 documents

dragon-dxw:
## Jira
https://national-archives.atlassian.net/browse/FCL-1252


script/clean_s3.py
```diff
@@ -0,0 +1,336 @@
...
+from datetime import datetime
+from typing import Any
+
+import boto3
+
+from caselawclient.models.utilities.aws import publish_documents
+from caselawclient.types import DocumentURIString
+
+# Configuration
+UNPUBLISHED_BUCKET = "unpublished-bucket-name"
```
dragon-dxw: Probably worth turning these strings into environment variables.

script/clean_s3.py
```diff
@@ -0,0 +1,336 @@
...
+PUBLISHED_BUCKET = "published-bucket-name"
+SNS_TOPIC_ARN = "arn:aws:sns:region:account:topic-arn"
+CHECK_INTERVAL_SECONDS = 30
+MAX_WAIT_MINUTES = 60
+DRY_RUN = True
+MAX_DOCUMENTS = None
+
+
+def extract_uri_from_key(key: str) -> str | None:
```
dragon-dxw: This feels weirdly off. If a key is d-a1b2c3/d-a1b2c3.docx, then the uri is /d-a1b2c3.docx which doesn't feel like it refers to anything real.

dragon-dxw: I think this needs to strip the extension too.

script/clean_s3.py
```diff
@@ -0,0 +1,336 @@
...
+                continue
+
+            try:
+                tag_response = s3_client.get_object_tagging(Bucket=bucket_name, Key=key)
+                tags = {tag["Key"]: tag["Value"] for tag in tag_response.get("TagSet", [])}
+
+                if "DOCUMENT_PROCESSOR_VERSION" not in tags:
+                    uris_needing_processing.add(uri)
+
+            except Exception as e:
```
dragon-dxw: bare except :(

dragon-dxw: Not convinced blindly processing the key for which we got a random unknown error is a great idea.

script/clean_s3.py
```diff
@@ -0,0 +1,336 @@
...
+                print(f"  Warning: Error checking {key}: {e}")
+                uris_needing_processing.add(uri)
+
+    uris_already_processed = len(all_uris) - len(uris_needing_processing)
+
+    print(f"  Found {len(all_uris)} published document URIs")
+    print(f"  Already processed (skipping): {uris_already_processed}")
+    print(f"  Need processing: {len(uris_needing_processing)}")
```
dragon-dxw: Add a "errored out" count

script/clean_s3.py
```diff
@@ -0,0 +1,336 @@
...
+    skipped_files = 0
+    skipped_unpublished = 0
+    queued_files = 0
+
+    for uri, data in uri_data.items():
+        if uri not in published_uris:
+            skipped_unpublished += sum(1 for _ in data["files"])
+            continue
+
+        has_docx = data["has_docx"]
```
dragon-dxw: smells like this whole thing wants to be object orientated

dragon-dxw (commented):
a

PR #1688, opened 2026-04-20
[FCLC-1498] Tidy up some XML manipulation methods and tests

jacksonj04:
Tidy up some rough edges now we have a clearer plan for in-app XML manipulation behaviours.

## Jira

FCLC-1498

## Checklist

### DIFF
--- a/src/caselawclient/models/documents/xml.py
+++ b/src/caselawclient/models/documents/xml.py
@@ -55,6 +55,18 @@ class XML:
     def get_xpath_nodes(self, xpath: str) -> list[Element]:
         return get_xpath_nodes(self.xml_as_tree, xpath, DEFAULT_NAMESPACES)
 
+    def get_single_xpath_node(self, xpath: str) -> Element:
+        """Return exactly one xpath node, and raise an exception if either it doesn't exist or exists multiple times."""
+
+        nodes = self.get_xpath_nodes(xpath)
+
+        if not nodes:
+            raise ValueError(f"No element found at xpath: {xpath}")
+        if len(nodes) > 1:
+            raise ValueError(f"XPath expression matches {len(nodes)} elements, expected exactly 1: {xpath}")
+
+        return nodes[0]
+
     def _modified(
         self,
         xslt: str,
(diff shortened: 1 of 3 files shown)


src/caselawclient/models/documents/xml.py
```diff
@@ -64,6 +55,18 @@ def get_xpath_match_strings(self, xpath: str) -> list[str]:
     def get_xpath_nodes(self, xpath: str) -> list[Element]:
         return get_xpath_nodes(self.xml_as_tree, xpath, DEFAULT_NAMESPACES)
 
+    def get_single_xpath_node(self, xpath: str) -> Element:
```
anthonyhashemi, line 58: What's the motivation for this new method?

src/caselawclient/models/documents/xml.py
```diff
@@ -42,15 +42,6 @@ def xml_as_string(self) -> str:
         """
         return str(etree.tostring(self.xml_as_tree).decode(encoding="utf-8"))
 
-    @property
-    def xml_as_bytes(self) -> bytes:
```
anthonyhashemi, line 46: I assume we removed this because it is unused across PUI and EUI now?

jacksonj04: Correct, any conversions to bytes (eg when sending multipart to MarkLogic) are handled at lower-level methods, so no need for it here. Helps keep the responsibility boundaries a bit cleaner.

src/caselawclient/models/documents/xml.py
jacksonj04: There were a few spots, most noticeably in testing, but also crucially in the `get_or_create_element` method which I forgot to add to the commit (😅), where we were getting the full list from `get_xpath_nodes`, checking the length was `1`, and then just using element `[0]`. This keeps all the logic in one place instead of repeating.

src/caselawclient/models/documents/xml.py
```diff
@@ -92,15 +95,7 @@ def get_or_create_element(self, parent_xpath: str, element_name: str, namespace:
...
-        parent_nodes = self.get_xpath_nodes(parent_xpath)
-        if not parent_nodes:
-            raise ValueError(f"No parent element found at xpath: {parent_xpath}")
-        if len(parent_nodes) > 1:
-            raise ValueError(
-                f"XPath expression matches {len(parent_nodes)} parent elements, expected exactly 1: {parent_xpath}"
-            )
-
-        parent = parent_nodes[0]
+        parent = self.get_single_xpath_node(parent_xpath)
```
anthonyhashemi, line 98: a-ha!

anthonyhashemi (approved):
nice

PR #1693, opened 2026-04-22
[FCLC-1824] fix(DocumentBody): remove cache on content_html

jacksonj04:
This is our chief suspect in some leaky memory behaviour; cache maintains a function-level cache of rendered HTML, but thanks to the inclusion of `self` is also effectively preventing every `DocumentBody` instance - potentially multiple per document (one per request, even for the same document) - from being garbage collected whilst not actually offering any cache benefits.

All user-facing rendered HTML should be cached at another layer already, so this is actually just wasting cycles and more crucially just chewing up memory until instances fall over.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog 1.0.0].
 
 ### Fix
 
+- **DocumentBody**: remove cache on content_html
 - test now correctly checks for existing value
 - fix bad test naming and redundant assertion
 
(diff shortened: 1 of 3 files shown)


anthonyhashemi (approved):
If this sorts our memory woes, this is the best value PR ever!

dragon-dxw (commented):
Some code may be written on the assumption this is cached; public detail_html build_context should be optimised a little but doesn't block this PR.

PR #1697, opened 2026-04-22
refactor: replace print debug statements with tiered logging

jacksonj04:
This gives us more flexibility in understanding what is going on whilst the code is running by dynamically adjusting log levels, rather than just blindly printing things to stdio.

### DIFF
--- a/src/caselawclient/models/documents/__init__.py
+++ b/src/caselawclient/models/documents/__init__.py
@@ -1,4 +1,5 @@
 import datetime
+import logging
 import os
 import warnings
 from functools import cached_property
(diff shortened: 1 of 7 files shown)


src/caselawclient/models/documents/__init__.py
```diff
@@ -440,10 +443,10 @@ def enrich(self, even_if_recent: bool = False, accept_failures: bool = False) ->
         Request enrichment of a document, if it's sensible to do so.
         """
         if not (even_if_recent) and self.enriched_recently:
-            print("Enrichment not requested as document was enriched recently")
+            logger.debug("Enrichment not requested as document was enriched recently")
```
dynamictulip: This feels like too low a log level - it's recording a decision rather than giving a glimpse into the runtime state so I think it should be an `info` rather than a `debug`

dynamictulip (approved):

jacksonj04:
I'm going to bypass the coverage requirements on the basis that if we didn't care about these logs before we don't care about them now. Any need for testing should surface itself.

PR #1727, opened 2026-04-27
Fix/marklogic client memory quick wins

jacksonj04:
## Jira

FCL-

## Checklist

- [ ] I have written tests to cover the new behaviour

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -302,7 +302,6 @@ class MarklogicApiClient:
             new_exception = new_error_class(
                 f"{e}. Response body:\n{response_body}",
             )
-            new_exception.response = response
             raise new_exception
 
     def _format_uri_for_marklogic(
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -376,8 +376,7 @@ class MarklogicApiClient:
         data: Optional[dict[str, Any]] = None,
     ) -> requests.Response:
         kwargs = self.prepare_request_kwargs(method, path, body, data)
-        self.session.headers = headers
-        response = self.session.request(method, **kwargs)
+        response = self.session.request(method, headers=headers, **kwargs)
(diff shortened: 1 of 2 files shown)


src/caselawclient/Client.py
```diff
@@ -75,6 +79,11 @@
 DEBUG: bool = bool(os.getenv("DEBUG", default=False))
 
 
+@lru_cache(maxsize=512)
```
dragon-dxw, line 82: This feels like running faster by trading memory, rather than the inverse. These should be limited in size and number, so probably fine, but runs counter to stated aim of PR.

src/caselawclient/Client.py
```diff
@@ -62,6 +64,8 @@
 # Requests timeouts: https://requests.readthedocs.io/en/latest/user/advanced/
 CONNECT_TIMEOUT = float(os.environ.get("CONNECT_TIMEOUT", "3.05"))
 READ_TIMEOUT = float(os.environ.get("READ_TIMEOUT", "10.0"))
+HTTP_POOL_CONNECTIONS = int(os.environ.get("MARKLOGIC_HTTP_POOL_CONNECTIONS", "10"))
```
dragon-dxw, line 67: It's potentially confusing that these are per-HTTP/HTTPS so actually could be double. In practice, it isn't going to be since we'll be using one or the other.

src/caselawclient/Client.py
```diff
@@ -204,6 +213,12 @@ def __init__(
         self.session = requests.Session()
```
dragon-dxw, line 213: I'm tempted to suggest that moving session config out of `__init__` into its own function may make sense.

src/caselawclient/Client.py
dragon-dxw, line 82: This cache will make things less responsive when developing.

dragon-dxw (approved):

PR #1746, opened 2026-05-07
[FCLC-1919] refactor: standardise how we represent etree._Element type

jacksonj04:
This was previously being described as both `caselawclient.xml_helpers.Element` (an alias for `lxml.etree._Element`) and directly as `lxml.etree._Element`. This is now standardised on just `caselawclient.xml_helpers.Element` across the codebase for clarity.

## Jira

FCLC-1919

## Checklist

### DIFF
--- a/src/caselawclient/models/documents/body.py
+++ b/src/caselawclient/models/documents/body.py
@@ -6,7 +6,6 @@ from typing import Optional
 
 import pytz
 from ds_caselaw_utils.types import CourtCode
-from lxml import etree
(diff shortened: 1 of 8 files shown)


src/caselawclient/models/documents/body.py
dragon-dxw, line 13: Is there a good reason to not 'from lxml import _Element as Element` rather than from xml_helpers?

anthonyhashemi (approved):

dragon-dxw (approved):

PR #1748, opened 2026-05-07
feat: add more logging around document publish

dynamictulip:
Adds more logs around document publish to track down where things might be falling over in ingestion

## Jira

FCL-1829

## Checklist

### DIFF
--- a/src/caselawclient/models/documents/__init__.py
+++ b/src/caselawclient/models/documents/__init__.py
@@ -453,8 +453,10 @@ class Document:
         except CannotEnrichUnenrichableDocument as e:
             if not accept_failures:
                 raise e
+            logger.warning("Enrichment failed for %s but proceeding anyway", self.uri, exc_info=True)
             return False
 
+        logger.info("Enrichment completed for %s", self.uri, exc_info=True)
         return True
 
     @cached_property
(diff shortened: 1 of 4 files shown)


src/caselawclient/models/documents/__init__.py
```diff
@@ -517,8 +523,10 @@ def publish(self) -> None:
         :raises CannotPublishUnpublishableDocument: This document has not passed the checks in `is_publishable`, and as
         such cannot be published.
         """
+        logger.info("Assert that document is publishable")
```
jacksonj04: This feels a bit more like `logger.debug` but I don't feel strongly.

src/caselawclient/models/utilities/aws.py
```diff
@@ -212,6 +213,7 @@ def announce_document_event(uri: DocumentURIString, status: str, enrich: bool =
             "StringValue": "1",
         }
 
+    logger.info("Announcing document event for %s with status %s and enrich %s", uri, status, enrich)
```
jacksonj04, line 216: This is probably another `debug`.

src/caselawclient/models/documents/__init__.py
dynamictulip: Agree :-)

src/caselawclient/models/utilities/aws.py
dynamictulip: This one I'd like to keep because it's useful to know that something should be about to change in the outside world

jacksonj04 (approved):

jacksonj04 (approved):

PR #1811, opened 2026-06-01
Relax 'can_enrich' so we can enrich bulk data

dragon-dxw:
Previously we blocked enrichment for bulk documents on the assumption that they were PDF-only and hence had no content: we think we can and should enrich bulk uploads which have content.

In future we should reparse them as well even if they have external data but that requires the external data to be roundtrippable to the Parser (and there's still a risk of data loss if that goes wrong)

## Jira

https://national-archives.atlassian.net/browse/FCLC-1968

## Checklist

- [ ] I have written tests to cover the new behaviour
- [X] I have created/updated method docstrings (if necessary)
- [X] I have considered if this is a breaking change

This is a change of behaviour but the current behaviour is no longer correct.

### DIFF
--- a/tests/models/documents/test_document_verbs.py
+++ b/tests/models/documents/test_document_verbs.py
@@ -216,7 +216,6 @@ class TestDocumentForceEnrich:
     @patch("caselawclient.models.documents.Document.can_enrich", new_callable=PropertyMock, return_value=False)
     def test_force_enrich_but_not_enrichable(self, mock_can_enrich, mock_announce_document_event, mock_api_client):
         document = Document(DocumentURIString("test/1234"), mock_api_client)
-        document.can_enrich = False
         with pytest.raises(CannotEnrichUnenrichableDocument):
             document.force_enrich()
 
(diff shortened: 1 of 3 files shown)


dragon-dxw:
This should fail at least one test that needs fixing. Tomorrow.

dynamictulip (commented):
We should add tests to ensure that the behaviour changed - can we hook into `TestDocumentEnrich`?

dynamictulip (approved):

tests/models/documents/test_document_verbs.py
```diff
@@ -228,6 +228,22 @@ def test_force_enrich_but_not_enrichable(self, mock_can_enrich, mock_announce_do
         mock_announce_document_event.assert_not_called()
 
 
+class TestDocumentCanEnrich:
+    @time_machine.travel(datetime.datetime(1955, 11, 5, 6))
```
dragon-dxw: I don't think this should be necessary (either time)

dynamictulip: I didn't know what this did so I copied it! Will try removing ;-)

dynamictulip: Fixed

PR #1821, opened 2026-06-04
FCLC-1990 | add version restore method to MarklogicApiClient 

dancs-dev:
Given a MarkLogic document with multiple versions
When the method is called to restore to a previous version
Then the method calls the MarkLogic xquery document-checkout-update-checkin

## Jira

Subtask `FCLC-1990` of `FCLC-1975`

## Checklist

### DIFF
--- a/src/caselawclient/Client.py
+++ b/src/caselawclient/Client.py
@@ -27,7 +27,7 @@ from caselawclient.models.documents import (
     DOCUMENT_COLLECTION_URI_PRESS_SUMMARY,
     Document,
 )
-from caselawclient.models.documents.versions import VersionAnnotation
+from caselawclient.models.documents.versions import VersionAnnotation, VersionType
 from caselawclient.models.judgments import Judgment
 from caselawclient.models.press_summaries import PressSummary
 from caselawclient.models.utilities import move
@@ -54,6 +54,7 @@ from .errors import (
     MarklogicResourceNotCheckedOutError,
     MarklogicResourceNotFoundError,
     MarklogicResourceUnmanagedError,
+    MarklogicResourceVersionInvalidError,
     MarklogicUnauthorizedError,
     MarklogicValidationFailedError,
 )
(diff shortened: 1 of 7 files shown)


src/caselawclient/models/documents/versions.py
```diff
@@ -26,6 +26,9 @@ class VersionType(Enum):
     EDIT = "edit"
     """ This version has been created as the result of a manual edit. """
 
+    RESTORE = "restore"
```
dancs-dev, line 29: If we go ahead with adding this new version type, we will need to have a think how this should be displayed in the EUI.

src/caselawclient/Client.py
```diff
@@ -1041,6 +1043,49 @@ def get_last_modified(self, judgment_uri: DocumentURIString) -> str:
...
+        Args:
+            document_uri: The URI of the document to restore.
+            version_number: The version number to restore.
+            annotation: An optional version annotation to store on the
+                restored version.
+
+        Returns:
+            The MarkLogic response from restoring the document.
+        """
+        # This only restores the document XML rather than properties (e.g.,
```
dragon-dxw: I'm concerned that we might want to rescue the previous annotation from the properties and replay it, but I think that might be because we're currently not really recording that metadata in a planned out way. I think this is good for now but I imagine we might change it in future.

src/caselawclient/Client.py
```diff
@@ -1041,6 +1043,49 @@ def get_last_modified(self, judgment_uri: DocumentURIString) -> str:
...
+        # previous TDR reference), as it seems like this is not currently stored
```
dancs-dev: I've checked and I believe the MarkLogic docs are in agreement with this comment: https://docs.marklogic.com/9.0/dls.documentAddProperties

>  Note that properties are not maintained in version history unless changes are also made to the content of the document.

dragon-dxw: wait, ... so according to that quote, properties *are* maintained in version history *if* changes are also made to the content? That feels odd.

dancs-dev: I think it may just be a subset of properties. I can't see a MarkLogic function that supports changing arbitrary properties as part of a content update. I'm guessing it refers to properties that you can set as part of a content update (such as annotation in the case of 
[document-checkout-update-checkin](https://docs.marklogic.com/dls:document-checkout-update-checkin)).

src/caselawclient/Client.py
dancs-dev: I think we may want to replay some of it, but I'd propose we do this as part of version restores from the `Document` model (to be added in next PR) rather than this lower-level API Client method.

However, I have updated the default annotation generated in this method so that it adds a `restored_from` to the payload.

dragon-dxw (approved):

PR #1850, opened 2026-06-12
feat: add restore_version method to Document

dancs-dev:
## Changes

- Added a new `restore_version` method to `Document` so document versions can be restored via the client API. Makes use of the lower-level method added in [1821](https://github.com/nationalarchives/ds-caselaw-custom-api-client/pull/1821).
- Included restore logic that re-applies metadata from the selected previous version, so restored documents keep consistent version/context metadata.
- Restores the S3 assets associated with the TDR submission of the version that is being restored to.
- Added tests for document verbs and document behavior around version restoration.

~~**Note:** Restoring assets in S3 will be done in a follow-up PR.~~

## Jira

FCLC-1991, FCLC-1992

## Checklist

- [X] I have written tests to cover the new behaviour - tests added
- [X] I have created/updated method docstrings (if necessary) - docstrings added
- [X] I have considered if this is a breaking change - adds rather than changes behaviour

### DIFF
--- a/src/caselawclient/models/documents/__init__.py
+++ b/src/caselawclient/models/documents/__init__.py
@@ -40,7 +40,13 @@ from caselawclient.models.utilities.aws import (
 from caselawclient.types import DocumentURIString, SuccessFailureMessageTuple, TDRMetadataDict
 
 from .body import DocumentBody
-from .exceptions import CannotEnrichUnenrichableDocument, CannotPublishUnpublishableDocument, DocumentNotSafeForDeletion
+from .exceptions import (
+    CannotEnrichUnenrichableDocument,
+    CannotPublishUnpublishableDocument,
+    CannotRestoreDocumentWithoutConsignmentReference,
+    CannotRestorePublishedDocument,
+    DocumentNotSafeForDeletion,
+)
 from .statuses import DOCUMENT_STATUS_HOLD, DOCUMENT_STATUS_IN_PROGRESS, DOCUMENT_STATUS_NEW, DOCUMENT_STATUS_PUBLISHED
 
 logger = logging.getLogger(__name__)
@@ -348,6 +354,32 @@ class Document:
                 + ", ".join(self.validation_failure_messages),
             )
 
+    def assert_is_restorable(self, version_number: int) -> None:
+        """Assert that the given version can be restored, raising if it cannot.
+
+        Args:
+            version_number: The version number that is to be restored.
+
+        Raises:
+            CannotRestorePublishedDocument: The document is currently published
+                and so cannot have a version restored, as the public version
+                could otherwise diverge from the restored draft content.
+            CannotRestoreDocumentWithoutConsignmentReference: No TDR consignment
+                reference is available to locate the document's S3 assets, so a
+                complete restore cannot be performed.
+        """
+        if self.is_published:
+            raise CannotRestorePublishedDocument(
+                f"{self.document_noun.capitalize()} {self.uri} cannot be restored because it is currently published.",
+            )
+
+        tre_metadata = self._get_restore_tre_metadata(version_number) or {}
+        if not tre_metadata.get("parameters", {}).get("TDR"):
+            raise CannotRestoreDocumentWithoutConsignmentReference(
+                f"{self.document_noun.capitalize()} {self.uri} cannot be restored to version {version_number} "
+                "because no TDR consignment reference is available to locate its assets.",
+            )
+
     @cached_property
     def first_published_datetime(self) -> Optional[datetime.datetime]:
         """
@@ -625,6 +657,24 @@ class Document:
 
         return metadata_source_version
 
+    def _get_restore_tre_metadata(self, version_number: int) -> Optional[dict[str, Any]]:
+        """Get the TRE metadata associated with a previous version.
+
+        Args:
+            version_number: Version number being restored.
+
+        Returns:
+            The `tre_raw_metadata` mapping from the metadata source version, or
+            `None` when there is no metadata source version or it carries no TRE
+            metadata.
+        """
+        metadata_source_version_document = self._get_restore_metadata_source_version(version_number)
+        if metadata_source_version_document is None:
+            return None
+
+        payload = metadata_source_version_document.structured_annotation.get("payload") or {}
+        return payload.get("tre_raw_metadata") or None
+
     def _get_version(self, version_number: int) -> Optional["Document"]:
         """Find a specific version from the document history.
 
@@ -687,10 +737,11 @@ class Document:
         """Restore a previous version of this document.
 
         Create a new version from a historical version while preserving version
-        history. Unpublish the document first, build the restore annotation from
-        the target version's structured annotation, attach action requester's
-        details to payload metadata, restore the corresponding S3 assets, and
-        reload the document body after restore.
+        history. Validate that the document is currently unpublished and the
+        required consignment metadata is available, build the restore annotation
+        from the target version's structured annotation, attach the action
+        requester's details to payload metadata, restore the corresponding S3
+        assets, and finally reload the document body and cached properties.
 
         Args:
             version_number: The version number to restore.
@@ -700,21 +751,34 @@ class Document:
 
         Raises:
             ValueError: If the specified version number is not found.
+            CannotRestorePublishedDocument: If the document is currently
+                published.
+            CannotRestoreDocumentWithoutConsignmentReference: If no TDR
+                consignment reference is available to locate the document's S3
+                assets.
         """
-
         restore_version_document = self._get_version(version_number)
-        metadata_source_version_document = self._get_restore_metadata_source_version(version_number)
 
-        # We need a version to restore from.
         if restore_version_document is None:
             raise ValueError(f"Version {version_number} not found in document history")
 
-        # Take the document out of the published state before restoring, so the
-        # public version cannot diverge from the restored draft content.
-        if self.is_published:
-            self.unpublish()
+        # Perform all necessary validation before updating the document:
+        self.assert_is_restorable(version_number)
+
+        # assert_is_restorable guarantees these are present.
+        tre_metadata = self._get_restore_tre_metadata(version_number) or {}
+        tdr_metadata = tre_metadata["parameters"]["TDR"]
+
+        consignment_reference_for_assets = tdr_metadata["Internal-Sender-Identifier"]
+
+        # Capture the source filename and images so S3 assets can be restored
+        # selectively, mirroring how the ingester stores them.
+        tre_payload = tre_metadata.get("parameters", {}).get("TRE", {}).get("payload", {})
+        source_filename_for_assets = tre_payload.get("filename")
+        image_filenames_for_assets: list[str] = tre_payload.get("images") or []
 
         restore_version_annotation = restore_version_document.structured_annotation
+
         # Cannot rely on .get("payload", {}) because payload may actually exist with a value of None.
         restore_version_annotation_payload = restore_version_annotation.get("payload")
         if restore_version_annotation_payload is None:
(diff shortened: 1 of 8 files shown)


src/caselawclient/models/documents/__init__.py
```diff
@@ -589,6 +600,145 @@ def delete(self) -> None:
...
+
+        restore_version_document = self._get_version(version_number)
+        metadata_source_version_document = self._get_restore_metadata_source_version(version_number)
+
+        # We need a version to restore from.
+        if restore_version_document is None:
+            raise ValueError(f"Version {version_number} not found in document history")
+
+        restore_version_annotation = restore_version_document.structured_annotation
+        # Cannot rely on .get("payload", {}) because payload may actually exist with a value of None.
```
anthonyhashemi, line 810: I wonder what cases result in this?

src/caselawclient/models/documents/__init__.py
```diff
@@ -589,6 +600,145 @@ def delete(self) -> None:
...
+            value=tdr_metadata["Internal-Sender-Identifier"],
+        )
+        self.api_client.set_property(
+            self.uri,
+            name="transfer-received-at",
+            value=tdr_metadata["Consignment-Completed-Datetime"],
+        )
+
+        # These have potentially been updated so remove from cache.
+        self.__dict__.pop("source_name", None)
```
anthonyhashemi, line 755: Apparently can do: `del self.source_name` etc.

src/caselawclient/models/documents/__init__.py
```diff
@@ -589,6 +600,145 @@ def delete(self) -> None:
...
+        version_document = None
+
+        for document in self.versions_as_documents:
+            if document.version_number == version_number:
+                version_document = document
+                break
+
+        return version_document
+
+    def _set_tdr_metadata(self, tdr_metadata: TDRMetadataDict) -> None:
```
anthonyhashemi, line 725: Nice! not for now, but we can eventually convert this to public and use it in ingester or use this to build a public ingest method in the apiclient

src/caselawclient/models/documents/__init__.py
```diff
@@ -589,6 +600,145 @@ def delete(self) -> None:
...
+        # Remove submitter as this could be misleading (they may not be the one who restored the version).
+        restore_version_annotation_payload.pop("submitter", None)
+        annotation = VersionAnnotation(
+            VersionType.RESTORE,
+            automated=automated,
+            message=f"Restored from version {version_number}",
+            payload=restore_version_annotation_payload,
+        )
+        annotation.set_calling_function("restore_version")
+        annotation.set_calling_agent(self.api_client.user_agent)
```
anthonyhashemi, line 833: for now I think this is fine, but once we move this to the EUI and think more about audit trails, we could consider adding an input of the specific user triggered it.

anthonyhashemi:
I haven't read through the tests yet but this generally looks really good. I am more concerned about the s3 side of things though, so may hold off approving this one until we have confidence in that one as otherwise this alone leaves us in a partial state which might be more problematic than not having it.

src/caselawclient/models/documents/__init__.py
```diff
@@ -589,6 +600,145 @@ def delete(self) -> None:
...
+
+            if tre_metadata := metadata_source_version_annotation_payload.get("tre_raw_metadata", {}):
+                if tdr_metadata := tre_metadata.get("parameters", {}).get("TDR", {}):
+                    self._set_tdr_metadata(tdr_metadata)
+
+                # Ensure restoring to a restored version can get required TDR metadata.
+                restore_version_annotation_payload["tre_raw_metadata"] = tre_metadata
+
+        # Remove submitter as this could be misleading (they may not be the one who restored the version).
+        restore_version_annotation_payload.pop("submitter", None)
```
dragon-dxw: Is this agreed somewhere? They're still the person who submitted the document originally.

dragon-dxw (approved):

src/caselawclient/models/documents/__init__.py
jacksonj04: We never really bottomed out exactly how tracking which user caused an action to happen would be recorded, since we dropped history stuff just before we got there.

I think we originally modelled it as having both immediate and proximal cause in the history, for storing eg "this revert was initialised by the EUI using API Client 12.34, _because_ User X requested it"

src/caselawclient/models/documents/__init__.py
jacksonj04: I'd err towards leaving submitter intact and recording the fact it was restored by someone somewhere else in the annotation. The shape of the annotation dict isn't really nailed down, so if we want to extend it to have some kind of new `action_requested_by` or similar value then we can do that safely.

src/caselawclient/models/documents/__init__.py
dancs-dev: I tested this out but unfortunately `del` raises an `AttributeError` if the value hasn't already been cached. We could wrap these in a `try/except` but that might be unnecessarily verbose.

src/caselawclient/models/documents/__init__.py
dancs-dev: I've left the submitter in but added a new optional `action_requested_by` field to the payload. As mentioned on Slack, we will probably want to tweak the EUI. Leaving the submitter in will result in them being named under the 'Changed by' column in the version history.

src/caselawclient/models/documents/__init__.py
anthonyhashemi: fair!

src/caselawclient/models/documents/__init__.py
dancs-dev: To be honest, I'm not actually sure if there are any judgments that don't have a payload, and this should be enforced by the typing system too. I encountered this error when creating unit tests, but this check is probably unnecessary. I'll adjust the comment to clarify, but would be tempted to leave the check in, in case there are any judgments that actually don't have a payload

src/caselawclient/models/documents/__init__.py
```diff
@@ -700,6 +702,11 @@ def restore_version(self, version_number: int, automated: bool = True) -> None:
         if restore_version_document is None:
             raise ValueError(f"Version {version_number} not found in document history")
 
+        # Take the document out of the published state before restoring, so the
+        # public version cannot diverge from the restored draft content.
+        if self.is_published:
```
anthonyhashemi: I think for now it would be better to check if published at the start of this method and if so, skip attempting to restore - similar to how we have checks on whether something can be published.

src/caselawclient/models/documents/__init__.py
```diff
@@ -733,6 +750,15 @@ def restore_version(self, version_number: int, automated: bool = True) -> None:
         annotation.set_calling_agent(self.api_client.user_agent)
 
         self.api_client.restore_document(self.uri, version_number, annotation)
+        # Without the consignment reference we cannot locate the tarball or know
```
anthonyhashemi: To start with, we should probably check this at the top as well and skip the whole restore as otherwise we will end up in weird partial states.

src/caselawclient/models/utilities/aws.py
```diff
@@ -164,6 +168,127 @@ def delete_non_targz_from_bucket(uri: DocumentURIString, bucket: str) -> None:
...
+        raise FileNotFoundError(
+            f"No matching consignment archive found in {bucket} for {uri} and consignment {consignment_reference}"
+        )
+
+    delete_non_targz_from_bucket(uri=uri, bucket=bucket)
+    archive_object = client.get_object(Bucket=bucket, Key=archive_key)
+    archive_bytes = archive_object["Body"].read()
+
+    with tarfile.open(fileobj=io.BytesIO(archive_bytes), mode="r:gz") as archive:
+        # Restore the source document under its canonical key (e.g. uksc_2023_1.docx).
```
anthonyhashemi, line 270: Once we move the ingester's save_files_to_s3 into apiclient then we will be able to extract out a shared method that can be share between this and that to ensure consistency

src/caselawclient/models/utilities/aws.py
```diff
@@ -164,6 +168,127 @@ def delete_non_targz_from_bucket(uri: DocumentURIString, bucket: str) -> None:
...
+        if source_filename:
+            source_extension = source_filename.rsplit(".", 1)[-1].lower()
+            source_destination_key = f"{uri}/{uri.replace('/', '_')}.{source_extension}"
+            if not _restore_archive_file_to_bucket(
+                archive, f"{consignment_reference}/{source_filename}", bucket, source_destination_key, client
+            ):
+                logger.warning("Source file %s not found in archive %s", source_filename, archive_key)
+
+        # Restore the parser log.
+        _restore_archive_file_to_bucket(
```
anthonyhashemi: For consistency, it would be good to have a similar log warning as we have for the source file itself and the images below

anthonyhashemi (changes requested):
This looks really good. I am requesting changes as per my comments so that calling this method doesn't leave us in an odd state in the unsupported edge case of an original tarball not existing for an older version of a record. Instead, it will error and we can investigate why we ended up in that unsupported state.

Also, whislt typing that out, it made me think about the case of an enriched document... how would this handle that case? The consignment ref for the corresponding unenriched version would have persisted once enriched and so can use that, right?

dancs-dev:
> This looks really good. I am requesting changes as per my comments so that calling this method doesn't leave us in an odd state in the unsupported edge case of an original tarball not existing for an older version of a record. Instead, it will error and we can investigate why we ended up in that unsupported state.
> 
> Also, whislt typing that out, it made me think about the case of an enriched document... how would this handle that case? The consignment ref for the corresponding unenriched version would have persisted once enriched and so can use that, right?

Thanks for reviewing this! I've implemented your suggestions. Totally agree, performing the initial validation and then moving the various updates to the end of the method sound very sensible. 

Yes, in the case of an enriched document, the consignment reference associated with the previous submission should be used.

src/caselawclient/models/documents/__init__.py
```diff
@@ -348,6 +354,32 @@ def assert_is_publishable(self) -> None:
                 + ", ".join(self.validation_failure_messages),
             )
 
+    def assert_is_restorable(self, version_number: int) -> None:
```
anthonyhashemi, line 385: This is a great abstraction!

tests/models/documents/test_document_verbs.py
```diff
@@ -1133,6 +1153,33 @@ def test_get_restore_metadata_source_version(
             assert result is not None
             assert result.version_number == expected_version_number
 
+    def test_get_restore_tre_metadata_returns_tre_raw_metadata_from_source(self, mock_api_client):
+        document = Document(DocumentURIString("test/1234"), mock_api_client)
+        source_version = _make_version_document(3, payload=_standard_tdr_payload())
+
+        with patch.object(Document, "_get_restore_metadata_source_version", return_value=source_version):
+            result = document._get_restore_tre_metadata(3)  # noqa: SLF001
```
anthonyhashemi: [nitpick]: I suggest testing the pathways in `_get_restore_tre_metadata` through tests of `restore_version` or `assert_is_restorable`, or considering whether _get_restore_tre_metadata should be a public method

anthonyhashemi (approved):
Great work

tests/models/documents/test_document_verbs.py
dancs-dev: Sounds sensible, I'm inclined to keep `_get_restore_tre_metadata` private but have moved its test coverage to `assert_is_restorable`. If this data was needed publicly, my feeling is we should have separate properties for each useful bit of the metadata.

I also removed the tests for `_get_version` as this is already tested through the `restore_version` tests.

anthonyhashemi (approved):

PR #1893, opened 2026-07-07
[FCLC-2014] Present all document metadata through a standardised interface in the API client

jacksonj04:
As the first step to making all our document metadata clean, beautiful and consistently implemented, create a new `document.metadata` submodule:

- Implements `Metadata`, `SingleMetadata` and `MultipleMetadata` base classes, which accept a `Document` as an argument. These are then inherited for:
- Implements classes (initially) which get the values for a `Document` for:
  - Name
  - Court
  - Jurisdiction
  - Date
  - Case number
  - Categories
- `Document` now has a list of metadata classes which are added to the `Document.metadata` dict, ready to be used for retrieval in future

## Jira

FCLC-2014

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,41 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog 1.0.0].
 
+## Unreleased
+
+### Feat
+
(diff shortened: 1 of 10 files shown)


jlhdxw (approved):
LGTM

PR #1897, opened 2026-07-09
[FCLC-2101] Clean up and standardise how we get metadata from XML body through to the new class-based metadata layer

jacksonj04:
Now that we have a new metadata object framework, we can move the behaviour of actually moving behaviour to the new subclasses.

For most of these, they rely (for the moment) on the contents of `document.body`. We move the XPath to a nice clean list in the `Body`, which still handles actually reading the XML. This is a candidate for better cleanup in future, but isn't really on the critical path here.

This is all a stepping stone to the metadata layer being able to handle multiple sources of facts.

## Jira

FCLC-2101

## Checklist

### DIFF
--- a/src/caselawclient/factories.py
+++ b/src/caselawclient/factories.py
@@ -18,9 +18,33 @@ from caselawclient.types import DocumentURIString
 
 T = TypeVar("T")
 
-DEFAULT_DOCUMENT_BODY_XML = """<akomaNtoso xmlns="http://docs.oasis-open.org/legaldocml/ns/akn/3.0" xmlns:uk="https://caselaw.nationalarchives.gov.uk/akn">
+
+def build_document_body_xml(
+    *,
+    name: str = "Judgment v Judgement",
+    court: str = "Court of Testing",
+    jurisdiction: str = "",
+    document_date_as_string: str | None = "2023-02-03",
+    case_number: str = "",
+) -> str:
+    frbr_date_line = (
+        f'                            <FRBRdate date="{document_date_as_string}"/>\n' if document_date_as_string else ""
+    )
+    return f"""<akomaNtoso xmlns="http://docs.oasis-open.org/legaldocml/ns/akn/3.0" xmlns:uk="https://caselaw.nationalarchives.gov.uk/akn">
             <judgment name="decision">
-                <meta/><header><p>Header contains text</p></header>
+                <meta>
+                    <identification>
+                        <FRBRWork>
+                            <FRBRname value="{name}"/>
+{frbr_date_line}                        </FRBRWork>
+                    </identification>
+                    <proprietary>
+                        <uk:court>{court}</uk:court>
+                        <uk:jurisdiction>{jurisdiction}</uk:jurisdiction>
+                        <uk:caseNumber>{case_number}</uk:caseNumber>
+                    </proprietary>
+                </meta>
+                <header><p>Header contains text</p></header>
                 <judgmentBody>
                 <decision>
                 <p>This is a document.</p>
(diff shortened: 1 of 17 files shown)


src/caselawclient/models/documents/metadata/types/date.py
```diff
@@ -9,6 +8,10 @@ class DateMetadata(SingleMetadata[datetime.date | None]):
     title = "Date"
     description = "The date of the document."
 
-    @cached_property
+    @property
+    def as_string(self) -> str:
+        return self.document.body.document_date_as_string
+
+    @property
```
dragon-dxw, line 17: Is this change from `value()` to `value` a breaking change? (Feels right though)

src/caselawclient/models/documents/body.py
```diff
@@ -19,6 +19,40 @@ class UnparsableDate(Warning):
     pass
 
 
+NAME_XPATH = "/akn:akomaNtoso/akn:*/akn:meta/akn:identification/akn:FRBRWork/akn:FRBRname/@value"
+COURT_XPATH = "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:court/text()"
+JURISDICTION_XPATH = "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:jurisdiction/text()"
+CATEGORIES_XPATH = "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:category"
```
dragon-dxw, line 27: This one feels different since it returns a node rather than a string. This feels right because it's `CATEGORIES` though.

src/caselawclient/models/documents/body.py
```diff
@@ -19,6 +19,40 @@ class UnparsableDate(Warning):
...
+CASE_NUMBER_XPATH = "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:caseNumber/text()"
+DATE_XPATH = "/akn:akomaNtoso/akn:*/akn:meta/akn:identification/akn:FRBRWork/akn:FRBRdate/@date"
+
+
+def categories_from_nodes(nodes: list[Element]) -> list[DocumentCategory]:
```
dragon-dxw, line 32: I think this'd benefit from a docstring.

src/caselawclient/models/documents/comparison.py
```diff
@@ -19,22 +19,22 @@ def __init__(
         # court, date, title
         self["court"] = AttributeComparison(
             label="court",
-            this_value=this_doc.court,
-            that_value=that_doc.court,
-            match=this_doc.court == that_doc.court,
+            this_value=this_doc.metadata["court"].value,  # type: ignore[attr-defined]
+            that_value=that_doc.metadata["court"].value,  # type: ignore[attr-defined]
```
dragon-dxw: The `attr-defined` type issue in production code does not spark joy, and suggests we're doing something smelly. Worth a look at?

src/caselawclient/models/documents/body.py
```diff
@@ -19,6 +19,40 @@ class UnparsableDate(Warning):
     pass
 
 
+NAME_XPATH = "/akn:akomaNtoso/akn:*/akn:meta/akn:identification/akn:FRBRWork/akn:FRBRname/@value"
+COURT_XPATH = "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:court/text()"
+JURISDICTION_XPATH = "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:jurisdiction/text()"
+CATEGORIES_XPATH = "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:category"
+CASE_NUMBER_XPATH = "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:caseNumber/text()"
```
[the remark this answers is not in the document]
dragon-dxw: This feels like the PR description is wrong.

src/caselawclient/factories.py
```diff
@@ -18,9 +18,33 @@
...
             <judgment name="decision">
-                <meta/><header><p>Header contains text</p></header>
+                <meta>
+                    <identification>
+                        <FRBRWork>
+                            <FRBRname value="{name}"/>
+{frbr_date_line}                        </FRBRWork>
+                    </identification>
+                    <proprietary>
+                        <uk:court>{court}</uk:court>
```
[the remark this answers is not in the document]
dragon-dxw: Not wrong, but it's only a test fixture.

dragon-dxw (approved):

src/caselawclient/factories.py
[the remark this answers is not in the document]
jacksonj04: Agreed, ignoring this until it becomes a problem.

src/caselawclient/models/documents/metadata/types/date.py
dragon-dxw: Nope. I'm wrong: the changes are actually "new property called date_as_string" and "value isn't cached" but the diff algorithm has unhelpfully got it wrong

PR #1919, opened 2026-07-14
[FCLC-2101] Convert Document.metadata to a more proper class

jacksonj04:
## Summary
- Replace the `DocumentMetadataRegistry` TypedDict with a real class
- Registry is still maintained via a structured `METADATA_TYPES` tuple
- Access is now strictly typed attributes: `document.metadata.name.value` (breaking)

## Test plan
- [ ] Document unit tests pass
- [ ] Spot-check comparison / force_reparse / validation helper paths
- [ ] Confirm deprecated Document attrs still route through registry attributes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog 1.0.0].
 
 ### BREAKING CHANGE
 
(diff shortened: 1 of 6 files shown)


jacksonj04:
Closing this as not worth the churn.

TypedDict already gives us strictly typed keyed access (`metadata["name"]`) and free enumeration (it's a dict at runtime). Moving to a class mainly buys dot-access, then forces us to reintroduce enumeration somehow — net benefit isn't there for a breaking change.

PR #1947, opened 2026-07-23
[FCLC-2175] Add ability to store and retrieve structured metadata in MarkLogic

jacksonj04:
We need to be able to store structured metadata for a document in MarkLogic, outside of the body proper. This PR introduces the ability to store and retrieve this structured metadata.

It supports the concepts of multiple conflicting claims from different sources, as well as falling back to the existing methods for getting values directly from the document body where no structured claims have been stored.

## Jira

FCLC-2175

## Checklist

### DIFF
--- a/src/caselawclient/models/documents/__init__.py
+++ b/src/caselawclient/models/documents/__init__.py
@@ -18,7 +18,13 @@ from caselawclient.errors import (
     OnlySupportedOnVersion,
 )
 from caselawclient.identifier_resolution import IdentifierResolutions
-from caselawclient.models.documents.metadata.registry import DocumentMetadataRegistry, MetadataAttributeKey
+from caselawclient.models.documents.metadata.fields.collection import MetadataFieldsCollection
+from caselawclient.models.documents.metadata.fields.exceptions import MetadataFieldValidationException
+from caselawclient.models.documents.metadata.fields.unpacker import unpack_all_metadata_fields_from_etree
+from caselawclient.models.documents.metadata.registry import (
+    DocumentMetadata,
+    MetadataAttributeKey,
+)
 from caselawclient.models.documents.metadata.types.case_number import CaseNumberMetadata
 from caselawclient.models.documents.metadata.types.categories import CategoriesMetadata
 from caselawclient.models.documents.metadata.types.court import CourtMetadata
@@ -142,7 +148,8 @@ class Document:
     Individual document classes should extend this list where necessary to validate document type-specific attributes.
     """
 
-    metadata: DocumentMetadataRegistry
+    metadata: DocumentMetadata
+    metadata_fields: MetadataFieldsCollection
 
     def __init__(self, uri: DocumentURIString, api_client: "MarklogicApiClient", search_query: Optional[str] = None):
         """
@@ -159,10 +166,11 @@ class Document:
 
         self._initialise_document_body(search_query=search_query)
         self._initialise_identifiers()
+        self._initialise_metadata_fields()
         self._initialise_metadata()
 
     def __repr__(self) -> str:
-        name = self.metadata["name"].value or "un-named"
+        name = self.metadata["title"].value or "un-named"
         return f"<{self.document_noun} {self.uri}: {name}>"
 
     def document_exists(self) -> bool:
@@ -200,16 +208,22 @@ class Document:
         identifiers_element_as_etree = self.api_client.get_property_as_node(self.uri, "identifiers")
         self.identifiers = unpack_all_identifiers_from_etree(identifiers_element_as_etree)
 
+    def _initialise_metadata_fields(self) -> None:
+        """Load this document's metadata_fields property from MarkLogic."""
+
+        metadata_fields_element = self.api_client.get_property_as_node(self.uri, "metadata_fields")
+        self.metadata_fields = unpack_all_metadata_fields_from_etree(metadata_fields_element)
+
     def _initialise_metadata(self) -> None:
         """Initialise all this document's metadata values."""
 
-        self.metadata = DocumentMetadataRegistry(
-            name=NameMetadata(self),
+        self.metadata = DocumentMetadata(
+            title=NameMetadata(self),
             court=CourtMetadata(self),
             jurisdiction=JurisdictionMetadata(self),
             date=DateMetadata(self),
             case_number=CaseNumberMetadata(self),
-            categories=CategoriesMetadata(self),
+            category=CategoriesMetadata(self),
         )
 
     @property
@@ -346,7 +360,7 @@ class Document:
 
     @cached_property
     def has_name(self) -> bool:
-        return bool(self.metadata["name"].value)
+        return bool(self.metadata["title"].value)
 
     @cached_property
     def has_valid_court(self) -> bool:
@@ -864,7 +878,7 @@ class Document:
 
         parser_instructions: ParserInstructionsDict = {
             "metadata": {
-                "name": self.metadata["name"].value or None,
+                "name": self.metadata["title"].value or None,
                 "cite": None,
                 "court": self.metadata["court"].value or None,
                 "date": checked_date,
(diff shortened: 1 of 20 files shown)


src/caselawclient/models/documents/metadata/base.py
```diff
@@ -11,10 +12,15 @@ class Metadata(ABC):
...
     title: ClassVar[str]
     description: ClassVar[str]
+    claim_name: ClassVar[str]
+    """Name attribute used in the MarkLogic ``metadata_fields`` property."""
 
     def __init__(self, document: "Document") -> None:
         self.document = document
 
+    def _resolve_claims(self) -> "ResolvedMetadataField":
+        return self.document.metadata_fields.resolve(self.claim_name)
```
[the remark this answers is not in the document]
jacksonj04: Done — `claim_name` is optional and defaults to `key` via `_resolve_claims()`.

src/caselawclient/models/documents/__init__.py
```diff
@@ -922,6 +932,10 @@ def save_identifiers(self) -> None:
                 "Unable to save identifiers; validation constraints not met: " + ", ".join(validations.messages)
             )
 
+    def save_metadata_fields(self) -> None:
+        """Persist the document's metadata_fields property to MarkLogic."""
+        self.api_client.set_property_as_node(self.uri, "metadata_fields", self.metadata_fields.as_etree)
```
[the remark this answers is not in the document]
jacksonj04: Done — `save_metadata_fields()` now runs `validate_ids_match_keys()` and raises `MetadataFieldValidationException` before writing to MarkLogic.

src/caselawclient/models/documents/metadata/fields/resolution.py
```diff
@@ -0,0 +1,52 @@
...
+        """All active claim values across every source (multi-value resolution)."""
+        return [claim.value for claim in self.claims]
+
+    @property
+    def winning_source(self) -> Optional[MetadataSource]:
+        """Source of the highest-precedence active claim, if any."""
+        active = self.claims
+        if not active:
+            return None
+        return max(active, key=lambda claim: claim.source.precedence).source
```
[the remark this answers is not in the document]
jacksonj04: Done — winner is the last active claim from the sorted list (source precedence, then latest timestamp, then id).

src/caselawclient/models/documents/metadata/types/categories.py
```diff
@@ -1,12 +1,41 @@
...
+
+    for parent, subcategories in children_map.items():
+        if parent in categories:
+            categories[parent].subcategories.extend(subcategories)
+
+    return [
+        categories[category_value.name]
+        for category_value in category_values
+        if category_value.parent is None and category_value.name in categories
+    ]
```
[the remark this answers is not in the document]
jacksonj04: Done — `document_categories_from_field_values` now de-duplicates by category name with first-seen order winning, so multi-source claims for the same name no longer produce duplicate tree entries.

src/caselawclient/models/documents/metadata/types/date.py
```diff
@@ -9,14 +10,35 @@ def date_as_string_from_value(value: datetime.date | None) -> str:
...
     title = "Date"
     description = "The date of the document."
 
     @property
     def value(self) -> datetime.date | None:
-        return self.document.body.document_date_as_date
+        resolved = self._resolve_claims()
+        if not resolved.has_any_claims:
+            return self.document.body.document_date_as_date
+        return date_from_metadata_field_value(resolved.value)
```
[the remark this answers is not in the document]
jacksonj04: Done — `test_date_unparsable_field_value_is_none` covers the structured-claim path (warns `UnparsableDate` and resolves to `None`).

src/caselawclient/models/documents/metadata/types/date.py
```diff
@@ -9,14 +10,35 @@ def date_as_string_from_value(value: datetime.date | None) -> str:
...
+    try:
+        return datetime.datetime.strptime(value, "%Y-%m-%d").date()
+    except ValueError:
+        # Import locally to avoid a circular dependency with DocumentBody.
+        from caselawclient.models.documents.body import UnparsableDate
+
+        warnings.warn(
+            f"Unparsable date encountered: {value}",
+            UnparsableDate,
+        )
```
jlhdxw, line 27: Does this get raised in alerting? I feel like we'd want to know if there are dates somehow being inserted that can't be parsed.

src/caselawclient/models/documents/metadata/types/name.py
```diff
@@ -3,9 +3,19 @@
 
 class NameMetadata(SingleMetadata[str]):
     key = "name"
+    claim_name = "title"
```
jlhdxw: How come we've added this for this one but not the others? key and claim_name are different?

src/caselawclient/models/documents/__init__.py
```diff
@@ -922,6 +933,15 @@ def save_identifiers(self) -> None:
                 "Unable to save identifiers; validation constraints not met: " + ", ".join(validations.messages)
             )
 
+    def save_metadata_fields(self) -> None:
+        """Persist the document's metadata_fields property to MarkLogic."""
+        validation_error = self.metadata_fields.validate_ids_match_keys()
```
jlhdxw: I wonder if exposing this or something like `validate_metadata_fields` would be useful so a client can handle their validation with the response easily?

tests/models/documents/test_document_metadata_fields.py
```diff
@@ -0,0 +1,309 @@
...
+        document = DocumentFactory.build(
+            api_client=mock_api_client,
+            body=DocumentBodyFactory.build(name="Body Name"),
+        )
+        document.metadata_fields.add(
+            MetadataField(
+                name="title",
+                value="Rejected Name",
+                source=MetadataSource.DOCUMENT,
+                id="abcdef01-2345-6789-abcd-ef0123456789",
```
jlhdxw: We have lots of Ids floating around in this file, with some that seem to be used to reference assertions, and some not. I wonder if it's worth randomly generating those that don't matter, to ensure they don't matter?

tests/models/documents/metadata/fields/test_metadata_fields.py
```diff
@@ -0,0 +1,390 @@
...
+                value=MetadataCategoryValue(name="Rejected"),
+                source=MetadataSource.EXTERNAL,
+                id="04a7b8c9-d0e1-4f2a-3b4c-5d6e7f8091a2",
+                timestamp=EARLY,
+                rejected=True,
+            )
+        )
+
+        values = collection.resolve("category").values
+        assert [value.name for value in values] == ["A", "B", "C"]  # type: ignore[union-attr]
```
jlhdxw: If we did `assert [value.name for value in values if value is not None] == ["A", "B", "C"]` I think we wouldn't need the ignore[union-attrs] comment

tests/models/documents/metadata/fields/test_metadata_fields.py
```diff
@@ -0,0 +1,390 @@
...
+    MetadataFieldRemovalNotAllowedException,
+)
+from caselawclient.models.documents.metadata.fields.field import MetadataCategoryValue, MetadataField
+from caselawclient.models.documents.metadata.fields.source import MetadataSource
+from caselawclient.models.documents.metadata.fields.unpacker import (
+    unpack_a_metadata_field_from_etree,
+    unpack_all_metadata_fields_from_etree,
+)
+
+EARLY = datetime(2025, 12, 17, 18, 0, 0, tzinfo=UTC)
```
jlhdxw: Shall we make these explicit `EARLY_TIMESTAMP` / `LATE_TIMESTAMP` when I saw these lower down I thought they were some inbuilt python magic!

src/caselawclient/models/documents/metadata/types/date.py
jacksonj04: No, _but_ this is consistent with existing behaviour. We should probably have another issue for "make sure warnings get properly logged"!

src/caselawclient/models/documents/metadata/types/name.py
jacksonj04: This is to help keep alignment with the metadata coming from the parser (currently enumerated somewhat painfully at https://github.com/nationalarchives/ds-caselaw-utils/blob/main/src/ds_caselaw_utils/schemas/parser-metadata/parser.schema.json#L203-L231).

_However_, since this is a whole new surface you're right in that we don't actually need this - we can just use `title` and make sure existing references to `name` proxy through.

jlhdxw (approved):
LGTM

PR #1978, opened 2026-07-27
[FCLC-2213] Add latest_published_datetime

jacksonj04:
Add `Document.latest_published_datetime` and set it on every publish (while keeping first-published only-if-unset).

## Jira

FCLC-2213

## Checklist


dynamictulip (approved):
LGTM.
Is there any way we can do a migration on the existing records?

PR #1982, opened 2026-07-27
refactor: make categories the canonical metadata key

jacksonj04:
## Summary
- Make `categories` the canonical `Document.metadata` / claim key (was `category`)
- Keep `category` as a deprecated facade alias with the usual warning
- Resolve both `name="categories"` and legacy `name="category"` claims so existing MarkLogic data keeps working
- Display title is now **Categories**

## Why
Multi-value metadata fields should use plural canonical keys (same convention as upcoming `judges`). Docs already document `Document.metadata["categories"]`.

## Test plan
- [x] `poetry run pytest tests/models/documents/test_document_metadata.py tests/models/documents/test_document_metadata_fields.py tests/models/documents/metadata/fields/test_metadata_fields.py`
- [ ] CI green
- [ ] Confirm EUI / other consumers using `metadata["category"]` still work via alias (with deprecation warning)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog 1.0.0].
 
+## Unreleased
+
+### Changed
+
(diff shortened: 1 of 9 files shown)


jacksonj04:
Superseded: folding into the unmerged factory-escape branch as a hard break (no category alias).

PR #1984, opened 2026-07-28
feat: add judges as structured document metadata

jacksonj04:
- Add `JudgesMetadata` with body fallback and metadata_fields claim resolution
- Expose `DocumentBody.judges` from Akoma Ntoso `judge` nodes
- Support editor claim add/suppress/restore, including materialising body names into DOCUMENT claims

Depends on the categories rename PR as base.

### DIFF
--- a/src/caselawclient/models/documents/body.py
+++ b/src/caselawclient/models/documents/body.py
@@ -69,13 +69,11 @@ def judges_from_nodes(nodes: list[Element]) -> list[str]:
     inside a ``judge`` node is flattened via ``itertext``.
     """
     judges: list[str] = []
-    seen: set[str] = set()
 
     for node in nodes:
         name = "".join(node.itertext()).strip()
-        if not name or name in seen:
+        if not name or name in judges:
             continue
-        seen.add(name)
         judges.append(name)
 
     return judges
--- a/src/caselawclient/models/documents/body.py
+++ b/src/caselawclient/models/documents/body.py
@@ -26,6 +26,7 @@ JURISDICTION_XPATH = "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:jurisdic
 CATEGORIES_XPATH = "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:category"
 CASE_NUMBER_XPATH = "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:caseNumber/text()"
 DATE_XPATH = "/akn:akomaNtoso/akn:*/akn:meta/akn:identification/akn:FRBRWork/akn:FRBRdate/@date"
+JUDGES_XPATH = "//akn:judge"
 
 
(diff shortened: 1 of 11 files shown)


src/caselawclient/models/documents/body.py
```diff
@@ -61,6 +62,25 @@ def categories_from_nodes(nodes: list[Element]) -> list[DocumentCategory]:
     ]
 
 
+def judges_from_nodes(nodes: list[Element]) -> list[str]:
+    """Extract unique judge display names from Akoma Ntoso ``judge`` nodes.
+
+    Names are returned in first-seen document order. Empty or whitespace-only
+    text is skipped; duplicate names are de-duplicated.
+    """
+    judges: list[str] = []
```
jlhdxw, line 71: What's the value in having a list for judges and a set for seen? Couldn't we just have either judges be a set, or not have the seen set at all and check in the judges list for existing names?

tests/models/documents/test_document_body.py
```diff
@@ -139,6 +139,45 @@ def test_categories(self, opening_tag, closing_tag):
...
+        )
+
+        assert body.judges == ["Judge One", "Judge Two"]
+
+    def test_judges_empty_when_absent(self):
+        body = DocumentBody(
+            b"""
+            <akomaNtoso xmlns="http://docs.oasis-open.org/legaldocml/ns/akn/3.0">
+                <judgment>
+                    <header><p>No judges here</p></header>
```
jlhdxw, line 173: Do we ever have judge nodes with html/characters that we need to handle?

jlhdxw (approved):
Looks good, minor comments/questions.

src/caselawclient/models/documents/body.py
```diff
@@ -27,6 +27,7 @@ class UnparsableDate(Warning):
 CATEGORIES_XPATH = "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:category"
 CASE_NUMBER_XPATH = "/akn:akomaNtoso/akn:*/akn:meta/akn:proprietary/uk:caseNumber/text()"
 DATE_XPATH = "/akn:akomaNtoso/akn:*/akn:meta/akn:identification/akn:FRBRWork/akn:FRBRdate/@date"
+JUDGES_XPATH = "//akn:judge"
```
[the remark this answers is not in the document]
jacksonj04: Fixed: scoped to `/akn:akomaNtoso/akn:*/akn:header//akn:judge`, with a regression test that body `<judge>` nodes are ignored.

tests/models/documents/test_document_body.py
jacksonj04: Yes — `judge` is mixed content, so `node.text` alone could truncate nested markup. Now using `itertext()` (covers nested inlines and decoded entities), with a test for that case.

src/caselawclient/models/documents/body.py
jacksonj04: Good call — with so few names the separate `seen` set was overkill. Simplified to just check membership on the list.

PR #1989, opened 2026-07-29
[FCLC-2225] Migrate to fully timezone-aware datetimes

jacksonj04:
- enforce UTC-aware datetimes at the strict `Client` and metadata parsing boundaries, rejecting naive inputs
- parse search result submission datetimes as UTC-aware values and normalize MarkLogic datetime properties to UTC
- enable `DTZ` in ruff and update affected tests and changelog output

## Jira

FCLC-2225

### DIFF
--- a/src/caselawclient/models/documents/metadata/fields/unpacker.py
+++ b/src/caselawclient/models/documents/metadata/fields/unpacker.py
@@ -7,6 +7,7 @@ from caselawclient.models.documents.metadata.fields.exceptions import (
 )
 from caselawclient.models.documents.metadata.fields.field import MetadataCategoryValue, MetadataField
 from caselawclient.models.documents.metadata.fields.source import MetadataSource
+from caselawclient.models.utilities.dates import require_aware_utc
 from caselawclient.xml_helpers import Element
 
 
@@ -56,7 +57,7 @@ def unpack_a_metadata_field_from_etree(metadata_xml: Element) -> MetadataField:
         ) from exc
 
     try:
-        timestamp = datetime.fromisoformat(timestamp_value)
+        timestamp = require_aware_utc(datetime.fromisoformat(timestamp_value), name="timestamp")
     except ValueError as exc:
         raise InvalidMetadataFieldXMLRepresentationException(
             f"Metadata field XML representation is not valid: unparsable timestamp '{timestamp_value}'"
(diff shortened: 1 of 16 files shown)


src/caselawclient/models/utilities/dates.py
```diff
@@ -3,6 +3,12 @@
 from dateutil.parser import isoparse
 
 
+def require_aware_utc(value: datetime, name: str = "datetime") -> datetime:
```
jlhdxw, line 6: Shall we add a little test for this?

src/caselawclient/models/documents/metadata/fields/unpacker.py
```diff
@@ -56,7 +57,7 @@ def unpack_a_metadata_field_from_etree(metadata_xml: Element) -> MetadataField:
         ) from exc
 
     try:
-        timestamp = datetime.fromisoformat(timestamp_value)
+        timestamp = require_aware_utc(datetime.fromisoformat(timestamp_value), name="timestamp")
```
jlhdxw, line 60: Do we need to add a test to cover `unpack_a_metadata_field_from_etree` with dates? I couldn't see where this is actually called so might be one of your test updates catches this?

src/caselawclient/models/utilities/dates.py
jacksonj04: Added direct coverage in `tests/models/utilities/test_util_dates.py` for both rejection of naive values and normalisation to UTC.

src/caselawclient/models/documents/metadata/fields/unpacker.py
jacksonj04: Happy-path unpacking was already covered via `TestMetadataFieldUnpacking`. I've added `test_unpack_naive_timestamp_raises` so the new aware-datetime guard is covered explicitly as well.

src/caselawclient/models/utilities/dates.py
```diff
@@ -1,9 +1,16 @@
+from datetime import UTC as DATETIME_UTC
 from datetime import datetime
 
 from dateutil.parser import isoparse
 from pytz import UTC, tzinfo
```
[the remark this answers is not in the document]
jacksonj04: Already addressed via the `remove-pytz` follow-up / rebase — this module now only uses stdlib `datetime.UTC`.

jlhdxw (approved):

PR #2036, opened 2026-08-06
[FCLC-2260] Materialise body metadata claims into metadata_fields

jacksonj04:
- Add idempotent materialisation of body-derived DOCUMENT metadata claims for all metadata fields
- Hash `LOGIC_VERSION` into `CURRENT_METADATA_MATERIALISATION_VERSION`, stored on documents as `latest_metadata_materialisation_version` to allow finding documents which don't have latest materialisation during maintenance
- `Document.materialise_metadata_claims()` stores claims and version, `Document.save()` runs it after saving XML.

MarkLogic change to add index: [range index on `latest_metadata_materialisation_version` (same branch name in `ds-caselaw-marklogic`).](https://github.com/nationalarchives/ds-caselaw-marklogic/pull/678)

## Jira
FCLC-2260

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog 1.0.0].
 
 ## Unreleased
 
+### Feat
+
+- materialise body-derived DOCUMENT metadata claims with per-field `LOGIC_VERSION` tracking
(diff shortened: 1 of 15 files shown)


jlhdxw (approved):
This looks good - we'll likely need to make some adjustments to the EUI view once this is released and that is updated with this version. I added some code to show the body claims from the document when there are no claims, so we can probably just remove that.

PR #2066, opened 2026-08-12
FCLC-2358: Per-type metadata pack/unpack; date claims as date

jacksonj04:
A bunch of improvements to how we serialise and deserialise metadata values between XML and memory, and some robustness improvements. This also better lays the ground for rich types like the upcoming party metadata, which has both name and role.

- Move claim value XML pack/unpack onto each metadata type, and introduce `PACK_VERSION` as a guard against us changing this in future
- Metadata values now handle normalisation (`normalise_metadata_field_value`)
- We actually validate a value type when constructing `MetadataField`
- **Breaking:** date metadata claim values are `datetime.date` in memory (wire format still `YYYY-MM-DD`)

# Jira

FCLC-2358

### DIFF
--- a/src/caselawclient/models/documents/metadata/base.py
+++ b/src/caselawclient/models/documents/metadata/base.py
@@ -1,13 +1,17 @@
 from abc import ABC, abstractmethod
 from collections.abc import Iterable
-from typing import TYPE_CHECKING, ClassVar, Generic, TypeVar
+from datetime import date
+from typing import TYPE_CHECKING, ClassVar, Generic, TypeVar, cast
 
 from caselawclient.models.documents.metadata.fields.field import (
-    MetadataCategoryValue,
+    MetadataDateValue,
     MetadataField,
     MetadataFieldValue,
+    MetadataStringValue,
 )
 from caselawclient.models.documents.metadata.fields.source import MetadataSource
+from caselawclient.models.documents.metadata.fields.unpack_helpers import stripped_element_text
+from caselawclient.xml_helpers import Element
 
 if TYPE_CHECKING:
     from caselawclient.models.documents import Document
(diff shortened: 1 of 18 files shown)


src/caselawclient/models/documents/metadata/base.py
```diff
@@ -75,6 +88,36 @@ class SingleMetadata(Metadata, Generic[T]):
...
+        """Like ``_string_value``, but suppressed/empty claims resolve to ``None``."""
+        resolved = self._resolve_claims()
+        if not resolved.has_any_claims:
+            return body
+        if resolved.value is None:
+            return None
+        self.validate_value(resolved.value)
+        return cast(MetadataStringValue, resolved.value).value or None
+
+    def _date_value(self, body: date | None) -> date | None:
```
jlhdxw: As we have a specific date class, is there any reason this should live outside that class?

tests/models/documents/metadata/fields/test_metadata_fields.py
```diff
@@ -42,6 +84,7 @@ def test_pack_scalar_value(self):
         assert element.get("source") == "document"
         assert element.get("timestamp") == EARLY_TIMESTAMP.isoformat()
         assert element.get("rejected") == "false"
+        assert element.get("pack_version") == "1"
```
jlhdxw: Will this fail when the value is updated? We have a comment on the setting of the value when the xml shape changes, but shall we add a note here to say we expect this to change when that also changes?

tests/models/documents/metadata/fields/test_metadata_fields.py
```diff
@@ -188,6 +236,128 @@ def test_unpack_empty_category_name_raises(self):
...
+    def test_pack_date_emits_iso_text(self):
+        field = MetadataField(
+            name="date",
+            value=MetadataDateValue(date(2023, 2, 3)),
+            source=MetadataSource.DOCUMENT,
+            id="9b2c8e4f-1a3d-4c5e-8f7a-6b0d9e2c1a34",
+            timestamp=EARLY_TIMESTAMP,
+        )
+        element = field.as_etree
+        assert element.get("pack_version") == "1"
```
jlhdxw: Sam here WRT the pack version adding a comment to make it clear this needs to be updated

tests/models/documents/metadata/fields/test_metadata_fields.py
jacksonj04: It will - adjusted so this just asserts that it matches the expected version from the class (ie it's being passed through properly).

jlhdxw (approved):

PR #2071, opened 2026-08-13
[FCLC-1677] Filter non-web image formats from HTML transforms

jacksonj04:
Some documents include WMF files, which can't be rendered by browsers, or by Pillow on Linux (which is used by Weasyprint to render fallback PDFs).

This PR modifies our XML to HTML XSLT transformation so that it filters out image formats which aren't web-friendly (whitelisting rather than blacklisting). These will now be silently dropped from output, rather than displaying as a broken image or causing a HTTP 500 during PDF rendering.

## Jira

FCLC-1677

### DIFF
--- a/src/caselawclient/models/documents/transforms/html.xsl
+++ b/src/caselawclient/models/documents/transforms/html.xsl
@@ -734,10 +734,14 @@
 </xsl:template>
 
 <xsl:template match="img">
-	<img loading="lazy">
-		<xsl:apply-templates select="@*" />
-		<xsl:apply-templates />
-	</img>
+	<!-- Only emit MDN common web image formats (omit WMF/EMF etc.).
+	     https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Formats/Image_types#common_image_file_types -->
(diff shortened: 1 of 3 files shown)


src/caselawclient/models/documents/transforms/html.xsl
```diff
@@ -734,10 +734,13 @@
 </xsl:template>
 
 <xsl:template match="img">
-	<img loading="lazy">
-		<xsl:apply-templates select="@*" />
-		<xsl:apply-templates />
-	</img>
+	<!-- Only emit browser-displayable image formats (omit WMF/EMF etc.). -->
+	<xsl:if test="matches(@src, '\.(gif|png|jpe?g|webp|svg)$', 'i')">
```
dancs-dev: **question:** do we want to add the other file types in this list? https://developer.mozilla.org/en-US/docs/Web/Media/Guides/Formats/Image_types#common_image_file_types

dancs-dev (approved):
Looks good to me, thanks for making the change

PR #2076, opened 2026-08-13
FCLC-2359: Idempotent MetadataFieldsCollection.add

jacksonj04:
- Idempotent `metadata_fields.add`: matching name/value/source is already present (including rejected; no revive); id collision / empty value / key–id mismatch raise
- Returns `MetadataFieldAddResult.ADDED` or `ALREADY_PRESENT`
- Collection errors subclass `MetadataFieldException`
- `__setitem__` routes through `add`; unpack dedupes via the same path
- Removed dead `validate_metadata_fields` / key–id save validation

## Jira

FCLC-2359

### DIFF
--- a/src/caselawclient/models/documents/metadata/fields/collection.py
+++ b/src/caselawclient/models/documents/metadata/fields/collection.py
@@ -1,20 +1,65 @@
+from enum import Enum
+
 from lxml import etree
 
 from caselawclient.models.documents.metadata.fields.exceptions import (
+    MetadataFieldEmptyValueException,
+    MetadataFieldIdCollisionException,
+    MetadataFieldKeyMismatchException,
     MetadataFieldRemovalNotAllowedException,
 )
 from caselawclient.models.documents.metadata.fields.field import MetadataField, MetadataFieldValue
 from caselawclient.models.documents.metadata.fields.resolution import ResolvedMetadataField
 from caselawclient.models.documents.metadata.fields.source import MetadataSource
-from caselawclient.types import SuccessFailureMessageTuple
 from caselawclient.xml_helpers import Element
+
+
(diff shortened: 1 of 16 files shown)


src/caselawclient/models/documents/metadata/fields/collection.py
```diff
@@ -1,20 +1,65 @@
...
+        Equivalence is ``name`` + ``value`` + ``source`` (``rejected`` is ignored,
+        so matching a rejected claim does not revive it). Empty values raise.
+        A reused id with a different payload raises
+        ``MetadataFieldIdCollisionException``.
+
+        Loading via unpack also goes through ``add``, so duplicate stored claims
+        with the same equivalence key are deduped on load.
+        """
+        if field.value.normalised() is None:
+            raise MetadataFieldEmptyValueException(f"Cannot add metadata claim '{field.name}': value is empty")
```
jlhdxw, line 46: Do we need a check for name too?

jacksonj04: It's _theoretically_ possible if someone hand-rolls a `MetadataField`, but in reality most of them are supposed to come via the metadata type classes so the name will be there regardless.

src/caselawclient/models/documents/metadata/fields/collection.py
```diff
@@ -1,20 +1,65 @@
...
+
```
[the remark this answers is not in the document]
jacksonj04: This is correct - empty values shouldn't exist in the wild, so if we come across one it's broken and we're into undefined behaviour (and failing hard is the correct response rather than guessing).

jlhdxw (approved):

PR #2160, opened 2026-09-02
FCLC-2355: Initialise document objects from XML and upsert save

jacksonj04:
Documents can now be constructed in memory from parser XML without an existing MarkLogic record, then persisted via a single `save()` call. This lays the foundation for simplifying how the Ingester performs updates without it having to replicate a bunch of knowledge about how `Document` objects behave.

- Add `Document.from_xml()` and `document_from_xml()` for in-memory construction; URIs mint as `d-{uuid4}` when not provided.
- Extend `Document.save()` to insert-or-update with `version_type` and `automated` parameters
- Add `is_persisted` and guard MarkLogic-backed calls until `save()` completes (`DocumentNotPersistedError`). This will eventually go away as we move more document behaviours to be in-memory.
- Refactor `DocumentFactory.build` to use the new construction path, so tests are using real `Document`s for more stuff instead of mocking.
- Deprecate `Client.insert_document_xml` / `update_document_xml` for lifecycle workflows in favour of `Document.save()`

**Breaking change:** `save()` on a URI that does not yet exist in MarkLogic now inserts the document instead of failing on update.

## Jira

FCLC-2355

### DIFF
--- a/src/caselawclient/models/documents/__init__.py
+++ b/src/caselawclient/models/documents/__init__.py
@@ -3,7 +3,7 @@ import logging
 import os
 import warnings
 from functools import cached_property
-from typing import TYPE_CHECKING, Any, ClassVar, Optional
+from typing import TYPE_CHECKING, Any, ClassVar, Optional, Self
 
 from ds_caselaw_utils import courts
 from ds_caselaw_utils.courts import CourtNotFoundException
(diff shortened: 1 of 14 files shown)


src/caselawclient/models/judgments.py
```diff
@@ -26,8 +27,10 @@ class Judgment(NeutralCitationMixin, Document):
     type_collection_name = "judgment"
     _default_reparse_document_type = "judgment"
 
-    def __init__(self, uri: DocumentURIString, *args: Any, **kwargs: Any) -> None:
-        super().__init__(self.document_noun, uri, *args, **kwargs)
+    def __init__(
+        self, uri: DocumentURIString, api_client: "MarklogicApiClient", search_query: str | None = None
```
jlhdxw, line 31: Is "MarklogicApiClient" supposed to be a string here, rather than referencing MarklogicApiClient ? is there a circular reference somewhere?

jacksonj04: Yes; it's only available when actually type checking otherwise we're loading a bundle of baggage, but putting it outside quotes gives us an undefined class error.

src/caselawclient/models/documents/__init__.py
```diff
@@ -167,6 +171,52 @@ def __init__(self, uri: DocumentURIString, api_client: "MarklogicApiClient", sea
...
+        return doc
+
+    @classmethod
+    def from_xml(
+        cls,
+        body: DocumentBody,
+        api_client: "MarklogicApiClient",
+        uri: DocumentURIString | None = None,
+    ) -> Self:
+        if cls is Document:
```
jlhdxw, line 212: Should we get checking the provided `cls` against the return value of `get_document_type_class` here? Is it possible for the cls to be passed in here to be wrong?

jlhdxw (approved):

